PR #50, opened 2021-10-04
dfu: update DFU sample to DFU v2 (with Golioth manifest)

mniestroj:


samples/dfu/src/flash.c
```diff
@@ -0,0 +1,174 @@
...
+/**
+ * Determines if the specified area of flash is completely unwritten.
+ *
+ * @note This is a copy of zephyr_img_mgmt_flash_check_empty() from mcumgr.
+ */
+static int flash_area_check_empty(const struct flash_area *fa,
+				  bool *out_empty)
+{
+	uint32_t data[16];
+	off_t addr;
```
0Grit, line 66: feel free to be verbose with variable names. eg "flashAreaAddress"

samples/dfu/src/flash.c
```diff
@@ -0,0 +1,174 @@
...
+	off_t end;
+	int bytes_to_read;
+	int rc;
```
0Grit, line 69: rc?

samples/dfu/src/flash.c
```diff
@@ -0,0 +1,174 @@
...
+	int i;
```
0Grit, line 70: "byteIndex", "byteIterator" ?

samples/dfu/src/flash.c
```diff
@@ -0,0 +1,174 @@
...
+		}
+
+		rc = flash_area_read(fa, addr, data, bytes_to_read);
+		if (rc != 0) {
+			flash_area_close(fa);
+			return rc;
+		}
+
+		for (i = 0; i < bytes_to_read / 4; i++) {
+			if (data[i] != 0xffffffff) {
```
0Grit, line 89: max size of?

samples/dfu/src/main.c
```diff
@@ -195,6 +77,71 @@ static int data_received(struct golioth_blockwise_observe_ctx *observe,
 	return 0;
 }
 
+static uint8_t *uri_consume_trailing_slash(uint8_t *uri, size_t *uri_len)
```
0Grit, line 80: Would this be useful as a standard helper function somewhere outside the example?

0Grit (commented):
Almost all comments on convention and encouragement to be verbose with variable names.

0Grit:
Since this isn't marked WIP I assume it's ready for merge. I'll wait on your comments back and will approve or continue commenting.

net/golioth/fw.c
```diff
@@ -0,0 +1,292 @@
...
+#include <qcbor/qcbor_spiffy_decode.h>
+#include <stdio.h>
+
+#include <logging/log.h>
+LOG_MODULE_DECLARE(golioth);
+
+#define GOLIOTH_FW_DOWNLOAD	".u"
+#define GOLIOTH_FW_DESIRED	".u/desired"
+
+#define MANIFEST_KEY_SEQUENCE_NUMBER		1
```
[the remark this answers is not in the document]
mniestroj: enums might be better here, indeed. They should provide better gdb experience. Will change that.

net/golioth/fw.c
```diff
@@ -0,0 +1,292 @@
...
+					&version_bufc);
+
+	QCBORDecode_GetTextStringInMapN(decode_ctx, COMPONENT_KEY_URI,
+					&uri_bufc);
+
+	qerr = QCBORDecode_GetError(decode_ctx);
+	if (qerr != QCBOR_SUCCESS) {
+		LOG_ERR("Failed to get version and URI: %d (%s)", qerr, qcbor_err_to_str(qerr));
+		err = -EINVAL;
+		goto exit_map;
```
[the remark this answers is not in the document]
mniestroj: Zephyr has inherited goto as a tool in error path handling from Linux kernel. I think it serves that purpose quite good. It is less error prone when there are some resources to be freed when using this convention as opposed to doing it manually in each "if (error) {" branch.

net/golioth/fw.c
```diff
@@ -0,0 +1,292 @@
...
+	}
+
+exit_root_map:
+	QCBORDecode_ExitMap(&decode_ctx);
+
+	return err;
+}
+
+int golioth_fw_observe_desired(struct golioth_client *client,
+			       struct coap_reply *reply,
```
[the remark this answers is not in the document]
mniestroj: As Zephyr inherits that from Linux kernel, here is an explanation why **not** to typedef structs: https://www.kernel.org/doc/html/v4.14/process/coding-style.html#typedefs

net/golioth/fw.c
```diff
@@ -0,0 +1,292 @@
...
+#define MANIFEST_KEY_HASH			2
+#define MANIFEST_KEY_COMPONENTS			3
+
+#define COMPONENT_KEY_PACKAGE			1
+#define COMPONENT_KEY_VERSION			2
+#define COMPONENT_KEY_HASH			3
+#define COMPONENT_KEY_SIZE			4
+#define COMPONENT_KEY_URI			5
+
+static int parse_component(QCBORDecodeContext *decode_ctx,
```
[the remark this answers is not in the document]
mniestroj: This is something inherited from Linux kernel as well. POSIX error codes are used to signal errors. This way it is easy to propagate error codes, as there is no need (in most cases) for mapping between one error code and the other. Additionally having single set of error codes allows developers to have a fixed mindset on how to handle errors.

include/net/golioth/fw.h
```diff
@@ -0,0 +1,78 @@
...
+ * @param version Pointer to version string, which will be updated by this
+ *                function
+ * @param version_len On input pointer to available space in version string, on
+ *                    output actual length of version string
+ * @param uri URI of the image, which will be updated by this function
+ * @param uri_len On input pointer to available space in URI string, on
+ *                output actual length of URI string
+ *
+ * @retval 0 On success
+ * @retval <0 On failure
```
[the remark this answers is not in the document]
mniestroj: Yes.

samples/dfu/src/flash.c
mniestroj: This variable is used in function called `flash_area_check_empty()`, so it already gives a context of what `addr` is about. I don't see a value in using `flash_area_` prefix.

Anyway, as noted in comment above function body, this is copy pasted from mcuboot code, with the same variable names. Or rather an older version of it, as I can see they have updated implementation in the meantime.

samples/dfu/src/flash.c
mniestroj: Return code. It is used sometimes in Zephyr, sometimes in mcumgr, sometimes in other west modules. I rarely use that myself and left that only because it was copy-pasted.

samples/dfu/src/flash.c
mniestroj: It has rather a meaning of erase flash, which is usually 0xffffffff. This is however not true all the time, which it seems they have addressed in mcumgr code already (by reading flash property describing what is the value after flash erase operation).

We can either update to newest implementation from mcumgr, or drop erased flash checking (which is an optimization).

samples/dfu/src/main.c
mniestroj: This implementation just removes first slash, so I wonder if function name describes it correctly. I am open to better name. Maybe `uri_strip_leading_slash()`?

samples/dfu/src/flash.c
mniestroj: Copy-pasted. But I like https://www.kernel.org/doc/html/v4.14/process/coding-style.html#naming rationale behind this. It is indeed really common to use `i` for simple counter and following same pattern everywhere results in high chances of seeing bugs (that do not follow that pattern).

net/golioth/fw.c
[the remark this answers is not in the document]
0Grit: I believe that modern IDE's now make it simple enough to find the context of what is behind the typedef.
I like typedefs because they force the user of the type to consider the implications of using that type somewhere it was not originally meant to be used.

While it doesn't 'enforce' stronger typing it at least encourages while also assisting with self documentation of the code.

samples/dfu/src/flash.c
0Grit: What sort of effort would updating MCU manager be?
I assume we'd do that in a separate PR? which would merge first?

samples/dfu/src/main.c
0Grit: The alternative naming you are proposing sounds better. 

I just wonder if other examples will end up stripping the leading slash and thus increase duplicated code.

net/golioth/fw.c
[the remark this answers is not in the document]
0Grit: I tend to return early in order to avoid the If nests but do end up having duplicate error handling logic sometimes.

Wish we just had exceptions.

net/golioth/fw.c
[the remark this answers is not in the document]
0Grit: C sucks.

net/golioth/fw.c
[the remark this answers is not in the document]
0Grit: Keep with convention for now. I'm not going to shake things up just yet.

samples/dfu/src/flash.c
0Grit: My brain doesn't like to assume. I also like that it provides some context when looking at the signature of the function and better self documentation.

It's approachable to new users of a codebase.

I'm not going to enforce but keep the newbs in mind.

samples/dfu/src/flash.c
0Grit: Your choice on whether to elaborate it to "returnCode" or leave as is.

samples/dfu/src/main.c
mniestroj: >The alternative naming you are proposing sounds better.

Updated.

>I just wonder if other examples will end up stripping the leading slash and thus increase duplicated code.

So far we didn't need that. If we'll need that at some point, then I'll create a uri_utils.h or whatever for such things.

samples/dfu/src/flash.c
mniestroj: Added new commit, that uses improved implementation of this function, from mcumgr.

samples/dfu/src/flash.c
mniestroj: I'll leave that as is, as currently this function is almost (95%, besides opening/closing flash area and whitespace differences) one-to-one equivalent of `zephyr_img_mgmt_flash_check_empty(uint8_t fa_id, bool *out_empty)`. It will be easier to update (as done minutes ago) when not changing variable names between the two.

alvarowolfx:
One thing that I noticed is that DFU is not checking/filtering for the `main` package specifically. That might break if someone creates a release with multiple packages.

0Grit:
Is 'main' package referenced anywhere in our spec of the functionality?

alvarowolfx:
Here is an example of a release manifest. There is this `"package":"main"` attribute with the package name. 
```
{"sequenceNumber":1633376389,"hash":"3967780c9c7e68acf9949183a241c519205a0c1831dbbd87f26f7e648299ac91","components":[{"package":"main","version":"1.0.4","hash":"ec0080c0b4327e7da02b39520ef37b3c54c12b571d03966df717b3f681152381","size":216880,"uri":"/.u/c/main@1.0.4","type":"mcuboot"}]}
```
Would be even better if we can pass that package name as a parameter on the application, so I can choose what is the host MCU package and later download separated packages ( like esp-at firmware, ML model, an asset for my device, or even a WASM module 👀  ) and save on flash. But I think that second part is a separate iteration of DFU, but I think will be nice to give that power to our users.

mniestroj:
In the first step we have decided with @0Grit to just allow single component firmware. We'll need to think a bit how to solve this multi component case, as we need to keep in mind resource constaints. I guess we need some design with application specific callback, that will be executed for each component that is found in the manifest, or alternatively for the whole array. That way application can decide to download some component or ignore it (for one or the other reason).

Anyway, I have prepared a commit that does check package name: https://github.com/golioth/zephyr-sdk/commit/a467238cd8a0d54ba0bc6c1a2596fe9bc34b46c5. I am not sure this is an improvement alone, as relying on single component instead of a component with package name "main" seems a bit more "future compatible".

net/golioth/fw.c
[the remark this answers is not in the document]
0Grit: I just saw your response to this now.  Did you end up making a change?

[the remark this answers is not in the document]
0Grit: Yes I se the change.

0Grit:
We are pre 1.0. I'm not worried too much about perfecting yet.

0Grit (approved):

PR #57, opened 2021-10-15
dfu: report state of firmware

mniestroj:
Introduced golioth_fw_report_state() function allows to report FW state
and result of download/update process.

Report firmware state in DFU sample with firmware download/update result.
Report current and target firmware version.


alvarowolfx (approved):
I tested here and worked as expected.

PR #51, opened 2021-10-04
samples: add ethernet overlays/configs

alvarowolfx:
Tested with the given boards/configurations:

| Board                                                                                                           | Processor   | Network  | Description                                                |
| --------------------------------------------------------------------------------------------------------------- | ----------- | -------- | ---------------------------------------------------------- |
| [ST Nucleo F746ZG](https://docs.zephyrproject.org/2.6.0/boards/arm/nucleo_f746zg/doc/index.html)                | STM32F746ZG | Ethernet | Built in Ethernet                                          |
| [WeAct Studio Black Pill V2.0](https://docs.zephyrproject.org/2.6.0/boards/arm/blackpill_f411ce/doc/index.html) | STM32F411CE | Ethernet | Wiznet W5500 module connected over SPI                     |
| [Particle Xenon](https://docs.zephyrproject.org/2.6.0/boards/arm/particle_xenon/doc/index.html)                 | nRF52840    | Ethernet | Particle Xenon + Ethernet Featherwing                      |

The ST Nucleo F746ZG was intentionally not added to the DFU sample because by default it doesn't have partitions defined on the Device Tree.

<a href="https://gitpod.io/#https://github.com/golioth/zephyr-sdk/pull/51"></a>

PR #76, opened 2021-11-09
samples: support nucleo_f746zg board

mniestroj:
This board provide ethernet interface, so extend sample code with
enabling of DHCPv4 client.

Add hal_st and hal_stm32 west modules, so that this board is built
correctly when this repo serves as west manifest repository.

Co-authored-by: Marcin Niestroj <m.niestroj@emb.dev>

Based on #51 with following changes:
- fix coding style
- include dhcpv4 header before using dhcpv4 api
- check if CONFIG_NET_DHCPV4 is enabled before starting DHCPv4 client
- do not use #ifdef around #include
- only add nucleo_f746zg board

<a href="https://gitpod.io/#https://github.com/golioth/zephyr-sdk/pull/76"></a>


0Grit:
These boards are not very available at the moment.

PR #78, opened 2021-11-11
samples: add support for esp32c3_devkitm platform

mniestroj:
As esp32c3_devkitm supports WiFi now, add overlay files which are really
a copy-paste of what we use for esp32 platform.

<a href="https://gitpod.io/#https://github.com/golioth/zephyr-sdk/pull/78"></a>


0Grit:
I think this should be one of our flagships yes.

0Grit (approved):

PR #79, opened 2021-11-11
system_client: check mbedTLS PSK length at build time and use it for runtime (config) buffer length

mniestroj:
Check PSK length at build time, so user will be notified early about PSK
being too big for use with mbedTLS.

So far we are fixed on using mbedTLS, so use configured PSK length as
the length of buffer for storing PSK for Golioth.

<a href="https://gitpod.io/#https://github.com/golioth/zephyr-sdk/pull/79"></a>

### DIFF
--- a/net/golioth/system_client.c
+++ b/net/golioth/system_client.c
@@ -328,7 +328,7 @@ void golioth_system_client_start(void)
  * credentials are stored. This means that we need to allocate memory for
  * credentials ourselves.
  */
-static uint8_t golioth_dtls_psk[64];
+static uint8_t golioth_dtls_psk[CONFIG_MBEDTLS_PSK_MAX_LEN];
 static size_t golioth_dtls_psk_len;
 static uint8_t golioth_dtls_psk_id[64];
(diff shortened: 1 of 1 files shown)


0Grit:
Does any documentation need updated as a result?

0Grit:
I think the way we may end up maintaining docs is for us to keep rough readme's internal to the sdk repository. Then those will be consumed by DevRel and or (Vance + Me)  in order to create guides for our documentation site.

mniestroj:
>Does any documentation need updated as a result?

I think that after this PR it is straightforward to figure out why 64 byte PSK will not work, because either user will use Kconfig and there will be build error right away with line of code specified what did not pass, or alternatively there will be error when trying to set too long PSK in which case user can check what is the buffer size.

If anything, I would rather add warning/error message in the second case (configuring PSK via settings subsystem), so that developers with logging enabled will see what is going wrong.

0Grit (approved):
Sorry forgot to complete this review.

PR #118, opened 2021-12-02
GitLab CI improvements

mniestroj:
TODO:
- [x] test if lightdb_led works with reduced heap size

<a href="https://gitpod.io/#https://github.com/golioth/zephyr-sdk/pull/118"></a>

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -97,9 +97,18 @@ pre-commit:
       done
       exit $CODE
 
-twister:
-  extends: .west-build
+.twister:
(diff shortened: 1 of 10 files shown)


0Grit:
@mniestroj  Thanks for 'spiking' this.  Once it's working can we separate the CI specific stuff into a separate PR or will the new CI changes fail what currently exists on main?

0Grit (approved):
Aproving.

Follow up discussion:

1. Can we create a single source of truth for our 'flagship' boards
2. Can we run CI on those boards based on their device tree?

PR #133, opened 2021-12-06
Move wifi.h into new app-support directory

0Grit:


0Grit:
@mniestroj asking for your review and I fully expect changes to be requested.

app-support/wifi/wifi.c
```diff
@@ -7,7 +7,7 @@
 #include <logging/log.h>
 LOG_MODULE_REGISTER(golioth_wifi, LOG_LEVEL_DBG);
 
-#include <net/golioth/wifi.h>
+#include <app-support/wifi.h>
```
mniestroj, line 10: So far we use `_` instead of `-` in directory names, I would prefer to keep that consistent and rename `app-support` to `app_support`.

net/CMakeLists.txt
```diff
@@ -1,2 +1 @@
-add_subdirectory(golioth)
-add_subdirectory(wifi)
+add_subdirectory(golioth)
```
mniestroj, line 1: There is "No newline at end of file" warning shown by GitHub. The common convention is to always add newlines at the end of text files. I've created #135, which will enforce that in CI.

app-support/Kconfig
```diff
@@ -0,0 +1 @@
+rsource "wifi/Kconfig"
```
mniestroj, line 1: There are options like:
```
config GOLIOTH_SAMPLE_WIFI
	bool "WiFi utilities for samples"
	depends on NET_L2_WIFI_MGMT
	help
	  Enable utilities for easy WiFi setup, mainly for use inside samples/.
```

So either we should place such options inside `sample_support` directory, or alternatively rename Kconfig options, so they would not mention "sample", but rather "app support" in their name.

mniestroj (changes requested):

app-support/Kconfig
0Grit: While these options are obviously tuned for our samples I'd like to imply they can be used by our users generically. In there applications.

Lets rename the Kconfig options to mention "app_support"

PR #147, opened 2022-01-07
Device matrix

mniestroj:


vitprajzler (commented):
The matrix looks great, thanks Marcin!

mniestroj:
@vitprajzler Is there some more information we want to store when testing? Or is there something that is not needed and can be dropped?

I wonder if we want to keep that Device Matrix as part of our SDK, for example in the README file as proposed in this PR. This would make it easier to update I think (at least to me) compared to 3rd party tool, such as Confluence. It would also benefit from storing the history of what has been tested and by who, all within git.

vitprajzler:
The only thing that would make the table easier to use is the Zephyr version. But since our SDK is released for a specific Zephyr version (or should be in the future), I don't think we need to add that / maintain that.

I think it's the right amount of information for now.

Definitely agree that keeping this somewhere accessible to our users is the better option than Confluence, and the automatic trace of who changed what and when.

In the long term, I think it would be worth keeping this in a simplified form in a JSON document and generate READMEs and other documents from it as part of docs (e.g. https://github.com/golioth/docs/blob/main/scripts/device-catalog/support.json), but for now just having this somewhere is good enough.

mniestroj:
Just updated with sha1 of the tip of main branch (this was changed after recent PR merges).

We can definitely use some json or yaml. However, then we would need to depend on generated version of this table as compared to main README.rst accessible directly from GitHub. No preference from my side, it just depends on where users will look for that information.

vitprajzler (approved):
Looks great!

And I agree, the "generated way" can wait.

PR #149, opened 2022-01-18
samples: dfu: reference specific revision of readme-espressif.md

mniestroj:
This (currently old) version of readme-espressif.md file specifies
offset of bootloader when flashing it on target, while the newest
version (on tip of main) does not reference any specific value and it is
not trivial to figure it out.

As this readme file might quickly evolve over time, let's stick to a
last "known to work" version of it.

<a href="https://gitpod.io/#https://github.com/golioth/zephyr-sdk/pull/149"></a>


alvarowolfx (approved):
LGTM

PR #151, opened 2022-01-25
samples: lightdb_led: convert from TinyCBOR to QCBOR

mniestroj:
QCBOR provides higher level APIs, so the general approach is to use
QCBOR instead of TinyCBOR. Convert lightdb_led sample to use QCBOR
library, so there will be single CBOR library used within our codebase.

<a href="https://gitpod.io/#https://github.com/golioth/zephyr-sdk/pull/151"></a>


vancefarren (approved):
Looks Good To Me

PR #154, opened 2022-01-26
net: golioth: fw: improve CBOR decoding error handling

mniestroj:
Call QCBORDecode_Finish() at the end of CBOR payload decoding, so that
potential string allocators get destructed properly. Use that function
to check for error codes instead of calling QCBORDecode_GetError().

Make sure to always call QCBORDecode_ExitMap() after
QCBORDecode_EnterMap(), even though it is not strictly required by QCBOR
implementation.

Make sure to always call QCBORDecode_ExitMap() after
QCBORDecode_EnterMap(), even though it is not strictly required by QCBOR
implementation.

Introduce a helper module, which will allow to convert from QCBOR error
codes to POSIX error codes. This will be particularly useful in error
handling paths in 'fw' and 'logging' modules.

Utilize introduced helper function to convert from QCBOR to
POSIX error codes. This will give higher level application code a better
knowledge about what went wrong underneath.

<a href="https://gitpod.io/#https://github.com/golioth/zephyr-sdk/pull/154"></a>


mniestroj:
There is a false positive from `checkpatch`, because of QCBOR library coding style.

vancefarren (approved):

PR #155, opened 2022-02-01
add delete method to lightdb api

vancefarren:


include/net/golioth.h
```diff
@@ -295,6 +295,21 @@ int golioth_lightdb_set(struct golioth_client *client, const uint8_t *path,
...
+ *
+ * @param client Client instance
+ * @param path Light DB resource path
+ * @param format Format of payload
+ *
+ * @retval 0 On success
+ * @retval <0 On failure
+ */
+int golioth_lightdb_delete(struct golioth_client *client, const uint8_t *path,
+			enum coap_content_format format);
```
mniestroj, line 311: `format` should not be needed, as there is no payload actually sent.

mniestroj (changes requested):

include/net/golioth.h
vancefarren: Ah, I thought that might be tied to the path specification in some way.  I'll test without.

net/golioth/golioth.c
```diff
@@ -498,6 +498,32 @@ int golioth_lightdb_set(struct golioth_client *client, const uint8_t *path,
...
+	}
+
+	err = coap_packet_append_option(&packet, COAP_OPTION_URI_PATH,
+					path, strlen(path));
+	if (err) {
+		LOG_ERR("Unable add uri path to packet");
+		return err;
+	}
+
+	return golioth_send_coap_payload(client, &packet, data, data_len);
```
mniestroj, line 524:
```suggestion
	return golioth_send_coap(client, &packet);
```

No reason to send any payload.

vancefarren: I didn't see that function. Thanks!

mniestroj:
@vancefarren Could you please cleanup that PR and just leave one commit with the required changes? There are also checkpatch and gitlint issues in CI that need to be addressed.

mniestroj:
@vancefarren ping

vitprajzler:
All checks have passed! :partying_face:

mniestroj (approved):

PR #161, opened 2022-02-04
net: golioth: fix decomposing path into URI-Path options

mniestroj:
Each path component (usually delimited by slash) should be encoded in
separate URI-Path option, instead of encoding whole path (includding
slashes) into single URI-Path.

Introduce coap_packet_append_uri_path_from_* helper functions:
 - *_from_string_range - gets pointer to 'begin' and 'end' of string,
 - *_from_string - gets pointer to string and its length,
 - *_from_stringz - gets pointer to NULL-terminated string.

Replace previous invocations of coap_packet_append_option(packet,
COAP_OPTION_URI_PATH, ...) with above functions, so that path is
properly decomposed into multiple URI-Path CoAP options.

<a href="https://gitpod.io/#https://github.com/golioth/zephyr-sdk/pull/161"></a>

### DIFF
--- a/net/golioth/CMakeLists.txt
+++ b/net/golioth/CMakeLists.txt
@@ -1,3 +1,3 @@
-zephyr_library_sources_ifdef(CONFIG_GOLIOTH golioth.c)
+zephyr_library_sources_ifdef(CONFIG_GOLIOTH coap_utils.c golioth.c)
 zephyr_library_sources_ifdef(CONFIG_GOLIOTH_FW fw.c)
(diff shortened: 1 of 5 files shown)


mniestroj:
It worked well so far, because there are probably fallbacks in both go-coap and Wireshark to handle single URI-Path. However previous implementation was not following https://datatracker.ietf.org/doc/html/rfc7252#section-6.4 and this is the reason why `aiocoap` server would not handle our requests properly.

PR #162, opened 2022-02-04
add new sensor sample for bh1749 and Golioth

vancefarren:


vitprajzler:
@vancefarren - please have a look at the checkpatch / code-style problems

vancefarren:
@mniestroj, what is the solution for an 'unrecognized platform' in the build check?

https://github.com/golioth/zephyr-sdk/runs/5068135399?check_suite_focus=true

mniestroj:
> @mniestroj, what is the solution for an 'unrecognized platform' in the build check?
> 
> https://github.com/golioth/zephyr-sdk/runs/5068135399?check_suite_focus=true

You can run twister locally and try to debug it. The issue is that `thingy91_nrf9160` platform is implemented in NCS repo, not in Zephyr (either mainline or Nordic fork). As a result when initializing repo with just Zephyr (without NCS), such board is not visible. Honestly I don't know how to fix that easily to make both vanilla Zephyr and NCS flavours testable, it is not trivial.

vitprajzler:
Good point @mniestroj, keeping the samples in this repo limited to one per device service sounds good to me. The per-device / per-sensor samples can quickly grow in number (and will need a different maintenance cycle), so having them in a separate repo would be a pragmatic step.

The support for the `thingy` board also points in the same direction.

Let's move this to the dedicated samples. 

@vancefarren , could you please create a new PR to the samples repo instead of this one?

vancefarren:
Yep, on my list

PR #165, opened 2022-02-08
Validate Thingy:91 with all samples

szczys:
This pull request collects the changes needed to make the Thingy:91 work with all Golioth samples.

<a href="https://gitpod.io/#https://github.com/golioth/zephyr-sdk/pull/165"></a>


samples/dfu/README.rst
```diff
@@ -211,14 +237,14 @@ Run following command on host PC to upload new firmware as artifact to Golioth:
...
-   $ goliothctl artifact create new.bin --version 1.2.3
+   $ goliothctl dfu artifact create new.bin --version 1.2.3
 
 Then create new release consisting of this single firmware and roll it out to
 all devices in a project:
 
 .. code-block:: console
 
-   $ goliothctl release --release-tags 1.2.3 --components main@1.2.3 --rollout true
+   $ goliothctl dfu release create --release-tags 1.2.3 --components main@1.2.3 --rollout true
```
mniestroj, line 247: Those lines should be fixed in separate commit. Bugfixes separately, new features separately.

samples/dfu/README.rst
```diff
@@ -7,44 +7,70 @@ Overview
...
 Requirements
 ************
 
 - Golioth credentials
 - Network connectivity
 
-Using nRF9160 Feather with nRF Connect SDK
-******************************************
+Using nRF9160 Feather or Thingy:91 with nRF Connect SDK
+*******************************************************
```
mniestroj, line 23: We do actually support nRF9160 DK as well, so maybe it is good to add it here as well (in separate commit of course).

samples/dfu/README.rst
```diff
@@ -7,44 +7,70 @@ Overview
 This DFU application demonstrates how to connect with Golioth and use Device
 Firmware Upgrade (DFU) procedure.
 
+The app is first built and flashed to the device which will then listen to
+Golioth Cloud for a new firmware binary to become available. A second binary is
+built with a new version number, registered as an artifact, and depolyed as a
```
mniestroj, line 12:
```suggestion
built with a new version number, registered as an artifact, and deployed as a
```

samples/dfu/README.rst
```diff
@@ -7,44 +7,70 @@ Overview
...
+release. The device will detect this change and automatically download and
+install.
+
```
mniestroj, line 15: This is an improvement of documentation and is unrelated to "feat: add support for Thingy:91 to dfu sample" commit title. One logical change == one commit. This commit should result at least with following commits:
- `samples: dfu: fix dfu artifact and release commands`
- some other documantation fixes
- `samples: dfu: improve high-level sample description`
- `samples: dfu: add support for Thingy:91 board`

Please remember to first push fixes, then improvements. Preferably bugfixes and/or documentation improvements should be in a separate PR, because they are not related directly to inclusion of Thingy:91 board and having everything in single PR will just block bugfixes from being included.

samples/dfu/README.rst
```diff
@@ -7,44 +7,70 @@ Overview
...
 - Golioth credentials
 - Network connectivity
 
-Using nRF9160 Feather with nRF Connect SDK
-******************************************
+Using nRF9160 Feather or Thingy:91 with nRF Connect SDK
+*******************************************************
+
+A full nRF9160 firmware `flashing overview`_ is available in the Golioth docs.
+Please use that resource to install the tools necessary to flash these devices.
```
mniestroj, line 26: What do you want to reference here? IMO Golioth docs should refer to SDK, not the other way.

samples/dfu/README.rst
```diff
@@ -7,44 +7,70 @@ Overview
...
 
-Build Zephyr sample application for nRF9160 Feather:
+by adding these lines to configuration file (e.g. ``prj.conf``):
+
+.. code-block:: cfg
+
+   CONFIG_GOLIOTH_SYSTEM_CLIENT_PSK_ID="my-psk-id"
+   CONFIG_GOLIOTH_SYSTEM_CLIENT_PSK="my-psk"
+
+Build Zephyr sample application:
```
mniestroj, line 40: This sounds like a generic doc improvement, so should be in separate commit. Hard to say if it is valid in current form, as there are too many changes as part of single commit.

samples/dfu/README.rst
```diff
@@ -7,44 +7,70 @@ Overview
...
 .. code-block:: console
 
-   $ west build -b circuitdojo_feather_nrf9160_ns samples/dfu
+   $ #Build for the Circuit Dojo nrf9160 Feather:
+   $ west build -b circuitdojo_feather_nrf9160_ns samples/dfu -p
+   $ #or build for the Thingy:91:
+   $ west build -b thingy91_nrf9160_ns samples/dfu -p
 
-Enter bootloader and use ``mcumgr`` to flash firmware:
+Enter bootloader mode and use ``mcumgr`` to flash firmware:
```
mniestroj, line 49: We are literally entering "bootloader", not some "bootloader mode" in application. I have mixed feeling whether adding "mode" here is and improvement or not. Just a doubt from my side, I won't block this though.

samples/dfu/README.rst
```diff
@@ -7,44 +7,70 @@ Overview
...
+follow those steps to create a new firmware release and roll it out to the
+device. **IMPORTANT:** use ``build/zephyr/app_update.bin`` instead of
+``new.bin`` in the first step when uploading the new artifact:
 
 .. code-block:: console
 
    $ goliothctl artifact create build/zephyr/app_update.bin --version 1.2.3
 
-See `nRF9160 Feather Programming and Debugging`_ for details.
-
```
mniestroj, line 47: Why don't we want to reference official documentation? Would be great to write down that reason in commit description.

samples/dfu/README.rst
```diff
@@ -124,7 +150,7 @@ Now build Zephyr sample application for nRF52840 DK:
 
 .. code-block:: console
 
-   $ west build -b nrf52840dk_nrf52840 samples/dfu
+   $ west build -b nrf52840dk_nrf52840 samples/dfu -p
```
mniestroj, line 153: Why adding `-p` everywhere? I don't think that Zephyr docs include that parameter, so why should we be different? If you will convince Zephyr maintainers to include that parameter in every command in docs, then I won't object to include that here. Let's be consistent.

samples/logging/README.rst
```diff
@@ -125,24 +125,40 @@ sample application (i.e., ``samples/logging``) and type:
...
 
 Enter bootloader and use ``mcumgr`` (or ``newtmgr``) to flash firmware:
 
 .. code-block:: console
 
-   $ mcumgr --conntype serial --connstring /dev/ttyUSB0,baudrate=1000000 build/zephyr/app_update.bin
+   $ #Flashing the Circuit Dojo nRF9160 Feather
+   $ newtmgr --conntype=serial --connstring='dev=/dev/ttyUSB0,baud=1000000' image upload build/zephyr/app_update.bin
+   $ #Flashing example for Thingy:91
+   $ newtmgr --conntype=serial --connstring='dev=/dev/ttyACM0,baud=115200' image upload build/zephyr/app_update.bin
```
mniestroj, line 161: There should be a reason to use `newtmgr` instead of `mcumgr`. Please put this change in separate commit and make sure to write down the reason of this change. Otherwise we will play ping-pong with switching between `mcumgr` and `newtmgr` if there is no real reason of using one or the other.

samples/logging/README.rst
```diff
@@ -125,24 +125,40 @@ sample application (i.e., ``samples/logging``) and type:
...
+
+Configure the build for the MCUboot bootloader by adding this line to the
+configuration file (e.g. ``prj.conf``):
+
+.. code-block:: cfg
+
+   CONFIG_BOOTLOADER_MCUBOOT=y
+
+On your host computer open a terminal window, locate the directory that contains
+the sample folder (i.e., ``~/zephyr-nrf/modules/lib/golioth``) and type:
```
mniestroj, line 145: It seems inconsistent to have that sentence only under `nRF9160 Feather or Thingy:91` subsection. In ALL cases we need to open a terminal change working directory.

> locate the directory that contains the sample folder (i.e., ``~/zephyr-nrf/modules/lib/golioth``)

This is the root directory of SDK repo. Sample folder is under `samples/<whatever>/`, so this is misleading now.

samples/lightdb_led/README.rst
```diff
@@ -213,4 +219,5 @@ as:
 .. _Networking with QEMU: https://docs.zephyrproject.org/latest/guides/networking/qemu_setup.html#networking-with-qemu
 .. _ESP32: https://docs.zephyrproject.org/latest/boards/xtensa/esp32/doc/index.html
 .. _AT Binary Lists: https://docs.espressif.com/projects/esp-at/en/latest/AT_Binary_Lists/index.html
+.. _flashing overview: https://docs.golioth.io/hardware/nrf91/quickstart/flash-firmware/flash-firmware
```
mniestroj, line 222: I think that there should be a one way direction of dependencies, not back and forth. In this specific case I think that `https://docs.golioth.io` should rather reference what is documented in SDK repo, not the other way.

samples/lightdb_led/README.rst
```diff
@@ -171,16 +187,6 @@ This is the output from the serial console:
...
-Monitor counter value
-=====================
-
-Device increments counter every 5s and updates ``/counter`` resource in Light DB
-with its value. Current value can be fetched using following command:
-
-.. code-block:: console
-
-   goliothctl lightdb get <device-id> /counter
-
```
mniestroj, line 183: Good catch. Deserves separate commit though. It is also a good idea to reference commit ID of the change that removed counter usage in code, so that it is easy to track down when documentation went out of sync. This is quite useful for maintenance of stable releases (which we don't do yet, but there are plans to start doing them).

mniestroj (changes requested):
Hard to review due to mixed bugfixes, improvements and new features, all as part of single commit.

Please consider to create a separate PR with bugfixes (and optionally obvious docs improvements). Then create another PR on top with the features you want to add (inclusion of Thingy:91).

Having `feat: ` in commit titles doesn't seem to have any real meaning, as those commits are including bugfixes anyway. This means that the title is a "lie". Anyway, we reference Zephyr commit/contribution guidelines (https://docs.zephyrproject.org/latest/contribute/index.html#commit-guidelines) and actually the first point says:
> A short and descriptive subject line that is less than 72 characters, followed by a blank line. The subject line must include a prefix that identifies the subsystem being changed, followed by a colon, and a short title, for example: doc: update wiki references to new site. (If you’re updating an existing file, you can use git log <filename> to see what developers used as the prefix for previous patches of this file.)

so `feat: ` is obviously not a good idea to be used in commit title.

samples/logging/README.rst
szczys: Okay, will do. When flashing the feather, the version of newtmgr that we recommend in our docs flashes much faster. We've normalized docs to use newtmgr for nRF91 examples.

samples/logging/README.rst
szczys: Good point, this would make more sense higher in the file to apply to all hardware compiling examples.

I made this change because our build instructions pass the `samples/<whatever`. If user is already in the directory for that sample, build will fail:
```
ERROR: source directory samples/lightdb/get does not exist
FATAL ERROR: refusing to proceed without --force due to above error
```
I think it makes more sense to new users to build from the same root directory every time.

samples/dfu/README.rst
szczys: This opens a larger question: do we add build command examples for every supported hardware?

samples/logging/README.rst
mniestroj: > I think it makes more sense to new users to build from the same root directory every time.

Makes sense!

samples/dfu/README.rst
mniestroj: > I think that information is too much to be included in the readme, 

I agree.

> but since we have documentation on it, we should make that known.

There are plenty of sites that already document that. I would rather stick closer to the source/vendor, i.e. reference Nordic (author of nrfjprog) and Jared (modified newtmgr with downstream patches) documentation instead.

Do we really want to break "Golioth docs ------> Golioth Zephyr SDK ------> Zephyr RTOS" one way relation? Isn't that going to make us pain later when synchronizing one to the other?

samples/dfu/README.rst
mniestroj: Good question. In general `samples/` in this repo should be just samples of what we implement in this repo, which is sample usage of lightdb, stream, dfu, etc. In the beginning we just had 3 platforms that we supported, so it made sense to give overlay files for each platform. Currently this list is much longer and the maintenance effort grows with each added platform, as in theory we should make sure that each commit still works on each platform.

samples/logging/README.rst
szczys: I started changing over the docs to use `mcumgr` instead of `newtmgr`. But then in testing I found that when an upload stalls (fairly frequently), you cannot stop the `mcumgr` process with CTRL-C; it must be killed from another terminal ([there's an open issue](https://github.com/apache/mynewt-mcumgr-cli/issues/21)).

This is a pretty bad user experience. On the one hand, `mcumgr` is an easier install process, but on the other hand `newtmgr` can be stopped with a keyboard interrupt.

Any thoughts on this?

szczys:
I've restructured this PR to isolate the changes related to adding Thingy:91 support.

This PR is rebased on #169 which should be merged first.

mniestroj:
@szczys Can you rebase this PR, so that merge conflicts are resolved?

One of the things I am missing now is build tests for this platform. We do have build tests for all other platforms, so that we don't accidentaly break them.

Additionally, we there is a [Device Matrix](https://github.com/golioth/zephyr-sdk#device-matrix) to be updated in main README file.

Last, but most important, are you commited to testing this platform regularly and maintaining support for it? IMO we already have too many platforms with no automated runtime tests, so I don't want to take that on me :)

szczys:
> Can you rebase this PR, so that merge conflicts are resolved?

Yes, done!

> One of the things I am missing now is build tests for this platform. We do have build tests for all other platforms, so that we don't accidentaly break them.

I haven't looked into these but with some guidance I'm happy to work on this.

> Additionally, we there is a [Device Matrix](https://github.com/golioth/zephyr-sdk#device-matrix) to be updated in main README file.

I can work on that. Is the SDK version the hash for this commit once it's merged? Where do I get golioth version info?

> Last, but most important, are you commited to testing this platform regularly and maintaining support for it? IMO we already have too many platforms with no automated runtime tests, so I don't want to take that on me :)

Sure? I agree that autoated runtime tests are important and we need to solve that. For now, I can run the tests on this device when needed.

szczys:
I made an update to the device matrix and added yaml entries in the samples for thingy91_nrf9160_ns.

szczys:
> However I think that we should assume that users understand west when working with SDK repo directly (as opposed to following docs repo).

It would be better assume the user understands enough to remove the `-p`. This is a possible tripping point when trying out Golioth. I was confused when I first encountered the need for a pristine build, especially since there is no `west clean` command (I'm used to typing `make clean`).

But, we can debate this another time. I'll remove these and push a commit so we can get this PR merged. Thanks!

samples/lightdb_led/README.rst
```diff
@@ -128,21 +128,27 @@ sample application (i.e., ``samples/lightdb_led``) and type:
...
+---------------------
 
 On your host computer open a terminal window, locate the source code of this
 sample application (i.e., ``samples/lightdb_led``) and type:
 
 .. code-block:: console
 
-   $ west build -b circuitdojo_feather_nrf9160_ns samples/lightdb_led
+   $ #Build for the Circuit Dojo nrf9160 Feather:
+   $ west build -b circuitdojo_feather_nrf9160_ns samples/lightdb_led -p
```
mniestroj: Could you please squash `-p` removal? There is no point in introducing it and removing right away in subsequent commit in the same PR.

szczys: Done! Please take a look when you have a chance. Thanks!

mniestroj (approved):

mniestroj:
@szczys Thanks a lot!

PR #180, opened 2022-04-07
samples: settings: improve shell comands to set/get setting

miguel91it:
In this PR we developed new command to the settings sample (get command) and included a '--json' flag to the shell commands (set/get) to print responses as json when setting or getting some setting.

<a href="https://gitpod.io/#https://github.com/golioth/zephyr-sdk/pull/180"></a>

### DIFF
--- a/samples/settings/src/settings.c
+++ b/samples/settings/src/settings.c
@@ -1,0 +1,195 @@
+/*
+ * Copyright (c) 2022 Golioth, Inc.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <zephyr.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <shell/shell.h>
+#include <sys/printk.h>
+#include <init.h>
+#include <ctype.h>
+
+#include <settings/settings.h>
+
+struct settings_read_callback_params {
+	const struct shell *shell_ptr;
+	bool value_found;
+	bool json_output;
+};
+
+static int cmd_settings_set(const struct shell *shell, size_t argc, char *argv[])
+{
+	bool json_output;
+
+	if (argc < 3) {
+		shell_fprintf(shell, SHELL_WARNING,
+			      "Wrong number of arguments\n");
+		shell_help(shell);
+		return -ENOEXEC;
+	}
+
+	json_output = false;
+	if (argc >= 4) {
+		if (strcmp(argv[3], "--json") == 0) {
+			json_output = true;
+		}
+	}
+
+	const char *name = argv[1];
+	const char *val = argv[2];
+	size_t val_len = strlen(val);
+
+	int err;
+
+	shell_fprintf(shell, SHELL_NORMAL, "Setting %s to %s\n", name, val);
+
+#ifdef CONFIG_SETTINGS_RUNTIME
+	err = settings_runtime_set(name, val, val_len);
+	if (err) {
+		if (json_output) {
+			shell_fprintf(shell, SHELL_VT100_COLOR_RED,
+				"{\"status\": \"failed\", "
+				"\"msg\": \"Failed to set runtime setting: %s:%s\"}\n", name, val);
+		} else {
+			shell_fprintf(shell, SHELL_ERROR,
+				      "Failed to set runtime setting: %s:%s\n", name, val);
+		}
+
+		return -ENOEXEC;
+	}
+#endif
+
+	err = settings_save_one(name, val, val_len);
+	if (err) {
+		if (json_output) {
+			shell_fprintf(shell, SHELL_VT100_COLOR_RED,
+				"{\"status\": \"failed\", "
+				"\"msg\": \"failed to save setting %s:%s\"}\n", name, val);
+		} else {
+			shell_fprintf(shell, SHELL_ERROR,
+				      "Failed to save setting %s:%s\n", name, val);
+		}
+
+		return -ENOEXEC;
+	}
+
+	if (json_output) {
+		shell_fprintf(shell, SHELL_VT100_COLOR_GREEN,
+			"{\"status\": \"success\", \"msg\": \"setting %s saved as %s\"}\n",
+			name, val);
+	} else {
+		shell_fprintf(shell, SHELL_NORMAL,
+			      "Setting %s saved as %s\n", name, val);
+	}
+
+	return 0;
+}
+
+static int settings_read_callback(const char *key,
+				  size_t len,
+				  settings_read_cb read_cb,
+				  void            *cb_arg,
+				  void            *param)
+{
+	ssize_t num_read_bytes = MIN(len, SETTINGS_MAX_VAL_LEN);
+	uint8_t buffer[num_read_bytes + 1];
+	struct settings_read_callback_params *params = param;
+
+	/* Process only the exact match and ignore descendants of the searched name */
+	if (settings_name_next(key, NULL) != 0) {
+		return 0;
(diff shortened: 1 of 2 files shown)


samples/settings/src/settings.c
```diff
@@ -0,0 +1,184 @@
+/*
+ * Copyright (c) 2021 Golioth, Inc.
```
mniestroj, line 2: It is 2022 already :)

samples/settings/boards/esp32.conf
```diff
@@ -0,0 +1,38 @@
...
+CONFIG_MBEDTLS_ECP_ALL_ENABLED=y
+
+
+#CONFIG_NET_L2_WIFI_SHELL=y
+#CONFIG_GOLIOTH_SYSTEM_SETTINGS=y
+#CONFIG_GOLIOTH_SAMPLE_WIFI_SETTINGS=y
+CONFIG_ESP32_WIFI_STA_AUTO=y
+CONFIG_GOLIOTH_SAMPLE_WIFI=n
+# CONFIG_GOLIOTH_SAMPLE_WIFI_SSID="Fim do Mundo"
+# CONFIG_GOLIOTH_SAMPLE_WIFI_PSK="19911996"
```
mniestroj, line 25: Please, no credentials here.

samples/settings/boards/esp32.conf
```diff
@@ -0,0 +1,38 @@
...
+
+CONFIG_MCUMGR=n
+CONFIG_MCUMGR_SMP_SHELL=n
+CONFIG_MCUMGR_CMD_CFG_MGMT=n
+CONFIG_MCUMGR_CMD_OS_MGMT=n
+
+CONFIG_GOLIOTH_SYSTEM_SERVER_HOST="coap.golioth.net"
```
mniestroj, line 32: Why do you change default value of `"coap.golioth.io"`?

samples/settings/boards/esp32.conf
```diff
@@ -0,0 +1,38 @@
+CONFIG_WIFI=y
```
mniestroj, line 1: Adding `esp32.conf` is separate logical change from "improve shell commands to set/get setting", so please create a separate commit and preferably a separate PR for that.

samples/settings/boards/esp32.conf
```diff
@@ -0,0 +1,38 @@
...
+CONFIG_MCUMGR_CMD_CFG_MGMT=n
+CONFIG_MCUMGR_CMD_OS_MGMT=n
+
+CONFIG_GOLIOTH_SYSTEM_SERVER_HOST="coap.golioth.net"
+
+CONFIG_FLASH=y
+CONFIG_FLASH_MAP=y
+CONFIG_NVS=y
+
+CONFIG_HWINFO=y
```
mniestroj, line 38: Why do we need this? We don't use that in other samples.

alvarowolfx: @mniestroj  I added the `CONFIG_HWINFO` to be able to check the device serial number or unique ID.I think is useful to add that to the other boards on the setting sample, so users can get a real hardware ID to be used to register the device on our platform.

vitprajzler: I think it's a good idea to extract the hardware ID from the device and register it on the cloud, rather than generating one in the cloud.

miguel91it: i kept it as is, right?

samples/settings/boards/esp32.conf
miguel91it: i removed from this PR and created a new one

mniestroj (changes requested):

samples/settings/src/settings.c
```diff
@@ -0,0 +1,185 @@
...
+				"{\"status\": \"failed\", \"msg\": \"setting not found\"}\n");
+		} else {
+			shell_error(shell, "Setting not found");
+		}
+	}
+
+	return 0;
+}
+
+SHELL_STATIC_SUBCMD_SET_CREATE(settings_commands,
```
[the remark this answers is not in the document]
mniestroj: Or even better improve the string provided in `SHELL_CMD_ARG()` macros. Unfortunately Zephyr shell commands are not consistent with how the help message should look like, but it is worth searching for a best candidate, so that users can learn how to use shell without leaving shell and readin comments that implement those shell commands.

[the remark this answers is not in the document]
ncmiller: > Unfortunately Zephyr shell commands are not consistent with how the help message should look like, but it is worth searching for a best candidate, so that users can learn how to use shell without leaving shell and readin comments that implement those shell commands.

@mniestroj - I'm not sure I understand the full context, but do you think this is the sort of thing that should be fixed in this PR, or is it okay to split this off into a separate backlog item?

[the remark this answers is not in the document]
mniestroj: We can split this off and improve in next PR.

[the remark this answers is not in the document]
miguel91it: Anyway, I improved the help string to help users understand without leaving the shell

```
uart:~$ settings
settings - Settings commands
Subcommands:
  set  :set a module config
        usage:
        $ settings set <key> <value>
        example:
        $ settings set golioth/psk pre-shared-key-value
        $ seetings set golioth/psk-id psk-identity-value
  get  :get a module config value
        usage:
        $ settings get <key>
        example:
        $ settings get golioth/psk
        $ settings get golioth/psk-id

```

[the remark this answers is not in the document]
miguel91it: what do you think?

samples/settings/src/settings.c
```diff
@@ -0,0 +1,184 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+#include <shell/shell.h>
+#include <sys/printk.h>
+#include <init.h>
+#include <ctype.h>
+
+#include <settings/settings.h>
+
+bool json_output;
```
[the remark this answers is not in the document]
miguel91it: Defining this variable at the function scope i cannot access it from another function.

In my specific case, i can define a local scope for the bool_json at the `cmd_settings_set` function.

But for the other 2 functions, `cmd_settings_get` and `settings_read_callback`, i did not find a way to do this because i need to access the boolean in the `settings_read_callback` and this function is called by the first, `cmd_settings_get` as a callback in a third function and, therefore, i cannot pass the boolean to the second function because i don't have control over the callback.

If i'm wrong could you help with any suggestions about it? Maybe i'm not seeing anything easy that you can see.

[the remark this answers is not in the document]
ncmiller: One other potential problem is that `params ` may no longer be in scope when `settings_read_callback` gets called. So I'd recommend making the struct instance static:

```c
	static struct settings_read_callback_params params = {
		.shell_ptr = shell,
		.value_found = false,
                .json_output = json_output
	};
```

ncmiller (approved):
Code changes look good to me! CI is failing right now, so we'll have to fix that before this can merge. But I'll go ahead and pre-approve.

miguel91it:
Ok Nick and thank you so much for your help.

mniestroj (approved):

ncmiller:
Okay - CI is failing right now. It will be fixed once https://github.com/golioth/zephyr-sdk/pull/185 merges, but that probably won't happen until sometime tomorrow.. For now, I'm going to go ahead and merge this, since the failing test in CI is benign.

PR #185, opened 2022-04-27
limit supported platforms to ESP32C3, nRF52840DK, nRF9160DK and QEMU_x86

mniestroj:
There were more and more "officially" supported platforms over time.
This was great, as it proved that implementation is platform agnostic
and it was possible to use any of the supported boards to try out this
project. There was however more and more maintenance effort needed in
order to test all of those platforms, as all the testing is done
manually so far.

Remove support of the following platforms:
 * circuitdojo_feather_nrf9160_ns - it is easier to use
     nrf9160dk_nrf9160_ns for development purposes, due to built-in
     SWD probe,
 * esp32c3_devkitm - it has similar feature set as 'esp32' platform
     (most importantly built-in WiFi); support was added just as a proof
     of concept and there is no allocated resources to keep maintaining
     it,
 * qemu_cortex_m3 - feature set is on par with qemu_x86,
 * thingy91_nrf9160_ns - see circuitdojo_feather_nrf9160_ns above.

It is still possible to use removed platforms, but configurationn and
most importantly testing of them for each sample/application is now out
of scope of this repository.

Remove overlay configuration files with following command:
```
  $ git ls-files | grep \
      -e circuitdojo_feather_nrf9160_ns\.conf \
      -e esp32c3_devkitm\.conf \
      -e thingy91_nrf9160_ns\.conf \
      -e qemu_cortex_m3\.conf \
      | xargs rm
```
Remove same set of platforms from device matrix and GitLab CI tests.

Update README files of all samples, so that removed platforms are no
longer referenced there.

<a href="https://gitpod.io/#https://github.com/golioth/zephyr-sdk/pull/185"></a>


ncmiller (approved):
> +101 -1030

Feels good to delete all of that stuff!

ChrisGammell:
This looks good for now, especially as we figure out testing methodologies. Longer term, I think we should potentially have a second tier of testing, even if community contributed. I would favor speed and staying sync'd with Zephyr/NCS over super broad support. 

We can work with frontend team to possibly push alternative discussions/notes in the docs for platforms that have oddities that require further documentation (ie. if the thingy91 requires additional overlay to work with the base nrf9160 files, etc)

ncmiller (changes requested):
~Blocking this PR, pending a decision to select `esp32` as our official board target (instead of `esp32c3_devkitm`)~

Okay, it's been decided that `esp32` is the official board target. Now I'm blocking until the PR gets updated with this change.

mniestroj:
Changed esp32c3 back to esp32.

ncmiller (approved):
Looks good, thanks!

szczys (approved):
LGTM. Sorry about the extra work in changing over the ESP32 variant, thanks for that!

PR #172, opened 2022-02-23
bump Zephyr to 3.0.0

mniestroj:
Bump Zephyr version in west manifest to just released 3.0.0 version.

TODO:
- [x] Test each platform with each sample, by following README file and verifying the expected behavior (according to README) with real behavior. After positive verification, a checkmark next to the platform/sample can be added. In case of failure, please add a comment to this PR with description about what went wrong or alternatively try to debug this issue.
  - [x] test "nRF91 Development Kit" with:
    - [x] [dfu](https://github.com/golioth/zephyr-sdk/tree/bump-zephyr-3.0.0/samples/dfu) sample
    - [x] [hello](https://github.com/golioth/zephyr-sdk/tree/bump-zephyr-3.0.0/samples/hello) sample
    - [x] [hello_sporadic](https://github.com/golioth/zephyr-sdk/tree/bump-zephyr-3.0.0/samples/hello_sporadic) sample
    - [x] [lightdb get](https://github.com/golioth/zephyr-sdk/tree/bump-zephyr-3.0.0/samples/lightdb/get) sample
    - [x] [lightdb observe](https://github.com/golioth/zephyr-sdk/tree/bump-zephyr-3.0.0/samples/lightdb/observe) sample
    - [x] [lightdb set](https://github.com/golioth/zephyr-sdk/tree/bump-zephyr-3.0.0/samples/lightdb/set) sample
    - [x] [lightdb led](https://github.com/golioth/zephyr-sdk/tree/bump-zephyr-3.0.0/samples/lightdb_led) sample
    - [x] [lightdb stream](https://github.com/golioth/zephyr-sdk/tree/bump-zephyr-3.0.0/samples/lightdb_stream) sample
    - [x] [logging](https://github.com/golioth/zephyr-sdk/tree/bump-zephyr-3.0.0/samples/logging) sample
    - [x] [settings](https://github.com/golioth/zephyr-sdk/tree/bump-zephyr-3.0.0/samples/settings) sample
  - [x] test "ESP32-WROOM-32D" with:
    - [x] [dfu](https://github.com/golioth/zephyr-sdk/tree/bump-zephyr-3.0.0/samples/dfu) sample
    - [x] [hello](https://github.com/golioth/zephyr-sdk/tree/bump-zephyr-3.0.0/samples/hello) sample
    - [x] [hello_sporadic](https://github.com/golioth/zephyr-sdk/tree/bump-zephyr-3.0.0/samples/hello_sporadic) sample
[...]

PR #187, opened 2022-04-29
Remove mcumgr from settings sample, use shell instead

ncmiller:
JIRA: https://golioth.atlassian.net/browse/DSDK-211

(see commits for description)

### Testing

- [x] ESP32: verified golioth credentials can be set via shell, and successfully connects to Golioth cloud
- [x] nRF91: verified golioth credentials can be set via shell, and successfully connects to Golioth cloud
- [x] nRF52+ESP-AT: verified golioth credentials can be set via shell, and successfully connects to Golioth cloud

<a href="https://gitpod.io/#https://github.com/golioth/zephyr-sdk/pull/187"></a>


mniestroj:
We no longer depend on TinyCBOR after this commit, so we can remove it from west manifest (west.yml).

ncmiller:
> We no longer depend on TinyCBOR after this commit, so we can remove it from west manifest (west.yml).

What about `mgmt/mcumgr`? That still depends on tinycbor. @szczys expressed concern over removing mcumgr entirely because there are still samples in the `samples` repo that use it.

ncmiller:
> > > We no longer depend on TinyCBOR after this commit, so we can remove it from west manifest (west.yml).
> > 
> > 
> > What about mgmt/mcumgr? That still depends on tinycbor. @szczys expressed concern over removing mcumgr entirely because there are still samples in the samples repo that use it.
> 
> Ohh, good question. I would rather have mcumgr functionality in the tree together with a sample (for build testing) or do not have it at all. Having implementation without anything that actually uses it will lead to broken functionality sooner than later. If we would not have mcumgr used in settings sample, then in #190 we would not even notice that something has broken.
> 
> I don't care if we keep mcumgr in the tree or not, so it is @szczys call. Samples repo should specify which SDK revision to use anyway, as most samples won't build with SDK tip of main anyway.

I agree. Since there are no samples remaining in the SDK that depend on mcumgr (and I don't think it makes sense to add one), then support for it should be removed entirely from the SDK.

The samples repo will just have to be clear about which version of the SDK the sample is depending on. This is something they already have to manage, since samples like `openthread` require currently non-supported versions of NCS (1.9.1).

@szczys - are you okay with this proposed strategy? We would be removing mcumgr support entirely from the SDK. It may be resurrected in the future if/when it makes sense (e.g. if we stop supporting ESP32 on Zephyr, we might bring it back).

ncmiller:
> Are we moving these working examples (ie: Thingy91) to the samples repo before removing them from the SDK? Since this was tested to be working on several board it would be a shame to lose track of how to do it.

Support for the Thingy91 is still in the `samples` repo: https://github.com/golioth/samples/blob/main/settings_mcumgr/boards/thingy91_nrf9160_ns.conf

What needs to happen is that the samples READMEs get updated to specify the SDK version they work with (probably the `unsupported_boards` tag, for now).

szczys:
Ah, that make sense. Thanks for clarifying.

This change LGTM.

ncmiller:
@mniestroj - to avoid having you retest in #172, I'd like to merge this now.

I will create another PR (after v0.1.0 releases) to:

* Remove all traces of mcumgr from the SDK
* Remove TinyCBOR

Sound good?

samples/settings/boards/nrf52840dk_nrf52840.conf
```diff
@@ -5,8 +5,6 @@ CONFIG_WIFI_ESP_AT_MDM_RX_BUF_COUNT=40
 CONFIG_NET_L2_WIFI_SHELL=y
 CONFIG_GOLIOTH_SAMPLE_WIFI=y
 
-CONFIG_MCUMGR_SMP_BT=y
-CONFIG_MCUMGR_SMP_BT_AUTHEN=n
 CONFIG_BT_L2CAP_TX_MTU=252
 CONFIG_BT_BUF_ACL_RX_SIZE=256
```
mniestroj, line 9: Please remove
```
CONFIG_BT_L2CAP_TX_MTU=252
CONFIG_BT_BUF_ACL_RX_SIZE=256
```
as BT is no longer enabled (indirectly by `CONFIG_MCUMGR_SMP_BT=y`). Those probably generate warnings right now at cmake configuration stage.

mniestroj (commented):
Looks good, just minor comments about BT for nRF52840 DK.

mniestroj:
> I will create another PR (after v0.1.0 releases) to:
> 
> * Remove all traces of mcumgr from the SDK
> * Remove TinyCBOR
> 
> Sound good?

Ok!

mniestroj (approved):

PR #190, opened 2022-05-05
Bump Zephyr to tip of main

mniestroj:
Everything except DFU on ESP32 work. DFU on ESP32 does not work right now due to a bug in Zephyr resulting in failed chain-loading from mcuboot.

<a href="https://gitpod.io/#https://github.com/golioth/zephyr-sdk/pull/190"></a>


mniestroj:
TinyCBOR to ZCBOR conversion is no longer needed after #190 will be merged.

Issue #158, opened 2022-02-02
Update SDK to work with NCS 1.8.0


vitprajzler:
The currently supported version of NCS is 1.7.1. 

We should update our SDK to support the latest current version of NCS, 1.8.0.

Reported in #157

mniestroj:
NCS 1.8.0 is broken.

Related issue on Nordic DevZone: https://devzone.nordicsemi.com/f/nordic-q-a/84041/bug-changes-to-nrf91_sockets-c-breaking-native-tls-when-using-libraries
Commit with breaking changes: https://github.com/nrfconnect/sdk-nrf/commit/27e7ebd847222ba0c7dd923add62558ffe57c4aa

PR #193, opened 2022-05-05
Bump NCS to tip of main

mniestroj:
After commit "lib: nrf_modem: don't enforce socket priority over native
TLS" was merged into 'main' branch it is possible to configure priority
of Zephyr TLS implementation to be higher than offloaded nRF9160 TLS.
This makes it possible to select Zephyr TLS (mbedTLS) implementation at
build time, which was not possible from NCS 1.8.0.

Update also CI scripts to be able to build newest NCS.

Solves: #158

<a href="https://gitpod.io/#https://github.com/golioth/zephyr-sdk/pull/193"></a>

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -13,10 +13,10 @@ jobs:
     runs-on: ubuntu-latest
 
     container:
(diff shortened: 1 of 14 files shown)


mniestroj:
Already done in #210

PR #202, opened 2022-05-10
gitlab-ci: use west-ncs.yml for 'twister-ncs' job

mniestroj:
So far NCS' west manifest file was patched in order to include Golioth
SDK as west module. Use 'west-ncs.yml', so that west workspace
initialization is simplified and the correct version is not hardcoded
into CI script.

Depends on (and contains commits from): #201


ncmiller (approved):
Thanks for cleaning this up

PR #209, opened 2022-05-13
samples: dfu: use settings to store credentials by default

mniestroj:
Make some cleanup across sample code that is used for more than a single
sample. See commits for details.

So far Golioth credentials were baked into the firmware after configuring
those with Kconfig (when used with DFU). Start using settings subsystem 
to store them instead, so that single firmware can be used for multiple
devices.

Disable settings subsystem for qemu_x86 platform, as there is no persisent
storage configured with default configuration used.


ncmiller (approved):
Nice cleanup. I'm glad to see the wifi code move into `samples/common`.

PR #210, opened 2022-05-13
bump Zephyr and NCS to tip of main

mniestroj:
After commit "lib: nrf_modem: don't enforce socket priority over native
TLS" was merged into 'main' branch of NCS it is possible to configure priority
of Zephyr TLS implementation to be higher than offloaded nRF9160 TLS.
This makes it possible to select Zephyr TLS (mbedTLS) implementation at
build time, which was not possible from NCS 1.8.0.

In case of Zephyr there are numerous networking stack and esp-at driver 
bugfixes in main branch.

Resolves: #158
Closes: #190 #193


ncmiller (approved):
:rocket:

PR #214, opened 2022-05-19
net: golioth: fix golioth_sendmsg() return code check

mniestroj:
Return code was saved to 'ret' variable, but 'err' variable was checked in
the next like instead.

In fact number of sent bytes should be compared with the number returned
from underlying zsock_send() call. Do exactly that inside
golioth_sendmsg(), as that function contains number of total bytes that
should be sent. Change golioth_sendmsg() behavior to return 0 in case of
success and negative value in case of error.

Save value returned from golioth_sendmsg() to 'err', which by convention
can be 0 (success) or negative (error).

Reported-by: Nick Miller <nick@golioth.io>

### DIFF
--- a/net/golioth/golioth.c
+++ b/net/golioth/golioth.c
@@ -299,7 +299,13 @@ static int golioth_sendmsg(struct golioth_client *client,
 
 	free(data);
 
-	return ret;
+	if (ret < 0) {
+		return ret;
+	} else if (ret < len) {
+		return -EIO;
+	}
+
+	return 0;
 }
 
 int golioth_send_coap(struct golioth_client *client, struct coap_packet *packet)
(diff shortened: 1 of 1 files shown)


net/golioth/golioth.c
```diff
@@ -299,7 +299,11 @@ static int golioth_sendmsg(struct golioth_client *client,
 
 	free(data);
 
-	return ret;
+	if (ret < len) {
+		return -EIO;
```
ncmiller, line 303: Could return `ret` here instead, since that would pick up the `errno` (in case the send call set an error other than EIO).

Alternatively, could split it into two cases:

```c
if (ret < 0) {
    return ret;
} else if (ret < len) {
    return -EIO;
}
```

mniestroj: Implemented the "alternative" version. EIO will be returned in case not full DGRAM was sent at once (in theory Zephyr should not allow that, but I have doubts that is the case).

ncmiller (approved):

PR #217, opened 2022-06-07
samples: common: move wifi.h header to samples/common/

mniestroj:
WiFi helper module was moved from net/wifi/ to samples/common/ in
commit ee8859e3f279 ("Move net/wifi/ to samples/common/ directory").
However, header file was missed during that time. Move it now to
samples/common/.

Introduce CONFIG_GOLIOTH_SAMPLES_COMMON Kconfig option, which is selected
by all samples sharing common samples code. This will prevent out-of-tree
applications to see such header files by default.


ncmiller (approved):
Looks good. 

We'll have to make similar changes to the `samples` repo soon, since several of those samples include `wifi.h` (cc @szczys @ChrisGammell).

PR #219, opened 2022-06-07
treewide: prepend zephyr/ to all Zephyr include paths

mniestroj:
Make code compatible with CONFIG_LEGACY_INCLUDE_PATH=n.

Not all Zephyr modules are compatible with this change, which is why
we have to wait before merging. Right now it it hard to verify that all
headers were converted properly.


mniestroj:
Changes already merged as part of #240

PR #230, opened 2022-07-13
Add golioth_rpc feature

ncmiller:
The RPC feature is not yet deployed on the backend (I tested with local `golioth` instance on branch `feature/device-rpc`). Therefore, the RPC feature is disabled by default in Kconfig.


net/golioth/Kconfig
```diff
@@ -39,6 +39,12 @@ config GOLIOTH_FW
 	  Enable Golioth firmware management, which allows to ask for newest
 	  desired firmware and issue firmware download process.
 
+config GOLIOTH_RPC
+	bool "Remote Procedure Call"
+	default n
```
mniestroj, line 44: As in upstream Zephyr project, please remove `default n` as all options are disabled by default.

net/golioth/Kconfig
```diff
@@ -39,6 +39,13 @@ config GOLIOTH_FW
 	  Enable Golioth firmware management, which allows to ask for newest
 	  desired firmware and issue firmware download process.
 
+config GOLIOTH_RPC_MAX_NUM_METHODS
```
mniestroj, line 48: As this is the first commit, I thought that this is an option that is not used within the codebase. In fact, it is used starting from the second commit. It should be introduced together with `config GOLIOTH_RPC` and with dependency on `config GOLIOTH_RPC`, so that number of RPC methods would not be configurable with RPC being disabled.

net/golioth/rpc.c
```diff
@@ -0,0 +1,253 @@
...
+struct golioth_rpc {
+	const char *method;
+	golioth_rpc_cb_fn callback;
+	void *callback_arg;
+};
+
+static struct golioth_client *_client;
+static struct coap_reply _observe_reply;
+static struct golioth_rpc _rpcs[CONFIG_GOLIOTH_RPC_MAX_NUM_METHODS];
+static int _num_rpcs;
```
mniestroj, line 47: I think that those should be moved to `struct golioth_client`, as so far there was no restriction on using multiple `golioth_client` instances.

Additionally `_rpcs` and `_num_rpcs` should probably be protected with some mutex or something. Or if that is not the case, then maybe we should put restrictions on when (from which thread, which calllback) `golioth_rpc_register()` can be called, so that there are no race conditions between threads.

net/golioth/rpc.c
```diff
@@ -0,0 +1,253 @@
...
+
+		/**
+		 * Call callback while decode context is inside the params array
+		 * and encode context is inside the detail map.
+		 */
+		QCBORDecode_EnterArrayFromMapSZ(&decode_ctx, "params");
+		QCBOREncode_OpenMapInMap(&encode_ctx, "detail");
+		status_code = matching_rpc->callback(&decode_ctx,
+						     &encode_ctx,
+						     matching_rpc->callback_arg);
```
mniestroj, line 154: This means that method will be called synchronously. Not sure about first version (maybe it is enough for now), but in the future we definitely want possiblity to send a reply after some long processing. This is what Azure IoT Hub allow and my experience D-Bus (on Linux as a RPC service) is that finally it is a desired thing to have. Just pointing it out for now.

include/net/golioth.h
```diff
@@ -64,8 +82,11 @@ struct golioth_client {
 	int sock;
 
 	void (*on_connect)(struct golioth_client *client);
-	void (*on_message)(struct golioth_client *client,
-			   struct coap_packet *rx);
+	void (*on_message)(struct golioth_client *client, struct coap_packet *rx);
+
+	/* Storage for additional on_message callbacks */
+	struct golioth_message_callback_reg message_callbacks[GOLIOTH_MAX_NUM_MESSAGE_CALLBACKS];
+	size_t num_message_callbacks;
```
mniestroj, line 89: We willy rework that and hide `on_message()` with the API rework. But for now it looks reasonable the way it is here.

net/golioth/rpc.c
ncmiller: I'll add mutex protection.

net/golioth/rpc.c
ncmiller: Yeah, I was thinking for that case, it would be up to the user to schedule the long-running work on the work queue or another task.

Alvaro and I ran into this issue when implementing the `reboot` method, because we were trying to reboot in the callback, which resulted in the RPC response never being sent, so we ended up scheduling the reboot in a separate task. That was a bit of an exception, because that method would effectively never complete in a synchronous way. In that case we ended replying with something along the lines of "reboot initiated", but not waiting for the reboot to complete.

We could completely remove `enum golioth_rpc_status` and expect the user to call a function like `golioth_rpc_send_response` either in the callback, or later on whenever the method is finished. This could cover both the synchronous method execution (response sent in callback) and asynchronous method execution cases (response sent after callback, possibly in another task). The main downside is that the QCBORDecodeContext and QCBOREncodeContext will go out of scope once the callback returns, so the user would need to copy stuff out and manage that themselves.

I think for now, I'll leave the synchronous approach, but I'll add a new enumerator to `enum golioth_rpc_status` for `GOLIOTH_RPC_RESPONSE_DEFERRED` and another function for `golioth_rpc_send_response`. This will at least provide some kind of option for users that need deferred execution, while keeping the synchronous case simple.

ncmiller: If it's okay with you, I'm going to open a separate JIRA issue to add support for async RPC. 

I played around with a few ideas (none of which felt right) and realized it will be a substantial amount of code change.

mniestroj:
There are build failures, which got unnoticed by CI: https://github.com/golioth/golioth-zephyr-sdk/runs/7396608549?check_suite_focus=true

I've observed this problem before and I think this was related to `twister` problem in Zephyr 3.1.0, which always returns 0 exit code even in case of errors. Zephyr from tip of `main` already fixed this issue, if I remember correctly.

net/golioth/rpc.c
```diff
@@ -0,0 +1,262 @@
...
+ */
+
+LOG_MODULE_DECLARE(golioth);
+
+#define GOLIOTH_RPC_PATH ".rpc"
+#define GOLIOTH_RPC_STATUS_PATH ".rpc/status"
+#define GOLIOTH_RPC_MAX_RESPONSE_LEN 256
+
+static struct golioth_client *_client = GOLIOTH_SYSTEM_CLIENT_GET();
+K_MUTEX_DEFINE(_rpc_mutex);
```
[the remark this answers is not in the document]
mniestroj: See #233 as a possible solution (not tested).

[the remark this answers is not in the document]
ncmiller: Merged that into this branch and rebased (added you as co-author on https://github.com/golioth/golioth-zephyr-sdk/pull/230/commits/12e1f0a280466fdf8afb51088d5b1311d0f1f398).

include/net/golioth.h
```diff
@@ -7,6 +7,9 @@
 #ifndef GOLIOTH_INCLUDE_NET_GOLIOTH_H_
 #define GOLIOTH_INCLUDE_NET_GOLIOTH_H_
 
+#if defined(CONFIG_GOLIOTH_RPC)
+#include <net/golioth/rpc.h>
+#endif
```
mniestroj, line 12: Would prefer to always include headers, regardless of CONFIG_ options. This is the policy followed by Zephyr. The advantage of that is the usage of `IS_ENABLED()`:
```
if (IS_ENABLED(CONFIG_GOLIOTH_RPC)) {
    golioth_rpc_register();
}
```
in application code, as opposed to:
```
#if CONFIG_GOLIOTH_RPC)
    golioth_rpc_register();
#endif
```
The advantage is that it is C compiler that removes unused code, not C preprocessor. This makes code more maintainable and harder to introduce errors based on option being enabled or disabled.

ncmiller: :+1: I'll remove the preprocessor statements, and condition things with `IS_ENABLED`

To avoid the extra storage in `struct golioth_client` in case RPC is not enabled, what do you think about doing this in `golioth_rpc.h`?

```c
struct golioth_rpc {
#if defined(CONFIG_GOLIOTH_RPC)
	bool initialized;
	struct coap_reply observe_reply;
	struct golioth_rpc_method methods[CONFIG_GOLIOTH_RPC_MAX_NUM_METHODS];
	int num_methods;
#endif
};
```

mniestroj (approved):

include/net/golioth.h
mniestroj: Seems fine to me!

Issue #220, opened 2022-06-12
samples/lightdb_led : QCBORDecode_Finish always logs a warning because it expects QCBOR_SUCCESS


drewwestrick:
### Zephyr Commit hash

0a904d0c176acf67ab10285e42f4d1924bc5bb79

### Golioth SDK Commit Hash

52e40f8627e95913b0d9729f6f6eaf04b4c44cff

### Hardware Target

nrf9160dk_nrf9160_ns

### What host OS are you using?

Ubuntu 20.04.4 LTS in WSL2

### Current Behavior

When you run the **samples/lightdb_led** on the **NRF91DK** and you change a value in the lightdb (web or api) the LEDs change state but the console reports  ```<wrn> golioth_lightdb: Did not enter CBOR map correctly```.

Looking at **QCBORError**, **QCBOR_ERR_NO_MORE_ITEMS** states that "If parsing a CBOR stream this indicates the non-error end of the stream."

I believe the return error **QCBOR_ERR_NO_MORE_ITEMS** represents the expected state and suggest a change to the expected result from **QCBOR_SUCCESS** to **QCBOR_ERR_NO_MORE_ITEMS**.

### Expected Behavior

**QCBORDecode_Finish** returns **QCBOR_ERR_NO_MORE_ITEMS**

### Steps To Reproduce

1. Clone the golioth-zephyr-sdk
2. Add the PSK credentials
3. Build the samples/lightdb_led for the nrf9160dk_nrf9160_ns board
4. Flash the firmware
5. Reboot the device and check the uart console

### Logs, console output, or any kind of debug information

```shell
[00:07:35.432,769] <dbg> golioth_lightdb: golioth_on_message: Payload
                                          a4 61 31 f4 61 32 f4 61  33 f4 61 30 f4          |.a1.a2.a 3.a0.
[00:07:35.432,983] <wrn> golioth_lightdb: Failed to finish decoding: 33 (QCBOR_ERR_NO_MORE_ITEMS)
```

PR #222, opened 2022-06-12
QCBORDecode_Finish updated to expect QCBOR_ERR_NO_MORE_ITEMS

drewwestrick:
If approved, this should close #220.

Signed-off-by: Drew Westrick <drewwestrick@glassboard.com>

PR #232, opened 2022-07-16
samples: lightdb_led: reset QCBOR error after "no more items"

mniestroj:
While loop is expected to run until "no more items" condition is reached.
Clear that decoding error, so that further processing can rely on
QCBOR_SUCCESS code.

Fixes: #220


mniestroj:
Alternative to #222.

ncmiller (approved):

PR #234, opened 2022-07-21
Add golioth_settings feature

ncmiller:
This is a thin layer that interfaces with the Golioth Settings
service, handling the CoAP observation of ".c" and sending
the response to ".c/status".

User code has to register a callback which will process a
single setting, and the library will call the callback for
each setting received from Golioth.

Added example code to the existing "settings" sample instead
of creating a new one, to avoid growing the already-extensive
list of samples.

------

Tested with ESP32 and nRF52+ESP32 on `coap.golioth.dev`.


samples/settings/src/main.c
```diff
@@ -10,9 +10,38 @@ LOG_MODULE_REGISTER(golioth_hello, LOG_LEVEL_DBG);
...
+		}
+
+		/* This setting must be in range [1, 100], return an error if it's not */
+		if (value->f < 1.0f || value->f > 100.0f) {
+			return GOLIOTH_SETTINGS_VALUE_OUTSIDE_RANGE;
+		}
+
+		/* Setting has passed all checks, so apply it to the loop delay */
+		_loop_delay_s = (int32_t)value->f;
+		LOG_INF("Set loop delay to %d seconds", _loop_delay_s);
```
mniestroj, line 37: Would be nice to have `int` support in addition to `float`. `float` has implicit rounding problems with high (enough) numbers. The rule I try to follow is NOT to use `float` when they are not directly beneficial. They are great for signal processing for example, assuming that MCU has FPU. However, for generic code, using `float` is like asking for trouble in corner cases. Either some libc math functions are invoked, that pull a lot of flash area to the firmware footprint or there is a rounding problem (like not getting the same value that anticipating) or with FPU enabled (users want to offload hard work to HW, right?) thread stacks are much more utilized due to more registers being dumped on thread context switch. Altough it is nice to have float, I really don't think they should be promoted, especially in case like above (int counter).

Just a comment, this is a design issue, so not a blocking thing for this PR.

ncmiller: Yeah, I agree we should have an `int` type.

I initially tried to add support for it. One issue is that the server sends us a double for all numbers, so there's an extra bit of code needed to check if the double is close enough to an integer to be rounded to it before handing it off to the user. I realized when trying to do this that the minimal libc in Zephyr doesn't support some functions I would need to do this in `math.h`, such as `round` or `modf`, and I didn't really have a great way to implement that myself from scratch. And I didn't want to have to depend on the more complete `newlib` just to have these functions.

But I might give this another shot. I realize now that I can multiply the double by 1000000 to move the decimal place over by 6 places, then check to see whether it's near a multiple of 1000000, and use that information to convert to int or not.

Alternatively, we can add support for `int` in `golioth` backend, then we wouldn't have to convert in the SDK.

mniestroj: > Alternatively, we can add support for int in golioth backend, then we wouldn't have to convert in the SDK.

+1 on that. I don't like the idea of rounding `double` to `int` in SDK and hiding that from user.

ncmiller: Opened this ticket for the backend to encode `int`s as `int`s instead of doubles: https://golioth.atlassian.net/browse/GB-346

I won't do the double-to-int conversion. I'll add a `TODO` in code for now, to add support for `int` once that ticket is done.

ncmiller: Opened this ticket for the backend to encode `int`s as `int`s instead of doubles: https://golioth.atlassian.net/browse/GB-346

I won't do the double-to-int conversion. I'll add a `TODO` in code for now, to add support for `int` once that ticket is done.

mniestroj (approved):

PR #236, opened 2022-07-22
DTLS certificate based authentication

mniestroj:
This is still a work in progress and subject to cleanup.


mniestroj:
Depends on #240 (bump Zephyr to tip of main).

net/golioth/Kconfig
```diff
@@ -26,6 +25,51 @@ module = GOLIOTH
...
+	select MBEDTLS_KEY_EXCHANGE_ECDHE_ECDSA_ENABLED if MBEDTLS_BUILTIN
+	help
+	  Select mbedTLS dependencies for certificate based authentication.
+
+	  This is a convenience option when using mbedTLS as a library to make
+	  (D)TLS connection with Golioth cloud.
+
+config GOLIOTH_HOSTNAME_VERIFICATION
+	bool "Hostname verification"
+	default y if GOLIOTH_AUTH_METHOD_CERT || GOLIOTH_AUTH_METHOD_PSK_AND_CERT
```
ncmiller, line 70: `GOLIOTH_AUTH_METHOD_PSK_AND_CERT`

I don't see this anywhere else.

net/golioth/system_client.c
```diff
@@ -171,6 +189,51 @@ static int init_tls(void)
...
+				 ca_certificate,
+				 ARRAY_SIZE(ca_certificate));
+	if (err < 0) {
+		LOG_ERR("Failed to register CA cert: %d", err);
+		return err;
+	}
+
+	err = tls_credential_add(CONFIG_GOLIOTH_SYSTEM_CLIENT_CREDENTIALS_TAG,
+				 TLS_CREDENTIAL_SERVER_CERTIFICATE,
+				 tls_server_certificate,
```
ncmiller, line 207: I understand that Zephyr calls it `TLS_CREDENTIAL_SERVER_CERTIFICATE`, but isn't this actually the public client certificate that we send to the server? Just double-checking my understand.

ncmiller (commented):
Looks really good, nice work!

net/golioth/Kconfig
mniestroj: Good catch! That is a leftover from refactoring. Initially I wanted to have 3rd option of "psk+cert" at the same time with possibility to choose at runtime, e.g. depending on provisioned credentials or so. But it was too much effort compared to the value IMO. Maybe we could introduce such option in the future, but first it would make more sense to refactor (split) TLS credentials management.

net/golioth/system_client.c
mniestroj: This name is confusing! Yeah, it is public client cert. I'll rename `tls_server_certificate` array to `tls_client_crt` to make it more consistent with Kconfig options, default file names, etc.

mniestroj:
Removed enabling of "extended master secret" from this PR, as it is unrelated (cloud no longer requires extended master secret).

mniestroj:
I can check if this PR works now without bumping Zephyr. Extended Master Secret (which is removed now) was possibly the only thing that required new Zephyr.

ncmiller (approved):

PR #242, opened 2022-07-28
settings: allow label type as byte or text string

ncmiller:
The prior assertion assumed the setting label was
always type QCBOR_TYPE_BYTE_STRING.

However, when testing settings, there was a board crash
due to detecting the label as type QCBOR_TYPE_TEXT_STRING.

Updating the assertion to allow either type for the label.


ncmiller:
`checkpatch` fails because of QCBOR Camel Case (`Avoid CamelCase: <uLabelType>`), which is a false-positive from our perspective.

mniestroj (approved):

PR #243, opened 2022-07-29
samples: add test sample

ncmiller:
Adding a new sample named test which is intended to run
automated tests in CI.

Currently, it only supports nrf52840dk, and has a single test
to verify that the board can connect to Golioth.

Another workflow was added for the nrf9160dk, but is currently
disabled due to inconsistency in modem initialization which
can cause the test to fail. It will be re-enabled once the
initialization issue is fixed (JIRA: DSDK-334).


samples/test/src/main.c
```diff
@@ -0,0 +1,27 @@
...
+
+ZTEST_SUITE(test_golioth, NULL, NULL, NULL, NULL, NULL);
+K_SEM_DEFINE(_connected_sem, 0, 1);
+
+static void golioth_on_connect(struct golioth_client *client)
+{
+	k_sem_give(&_connected_sem);
+}
+
+ZTEST(test_golioth, test_connect)
```
mniestroj, line 22: Is there a reason to use ZTEST for that sample? Any advantages it gives in this context? So far I've seen ZTEST use only in unit/integration tests that were self-sufficient, i.e. without any external verification like the Python code in this sample.

ncmiller: The truthful answer is - that's the way I did it in the ESP-IDF SDK, and it worked out okay (using Unity test framework which is similar to ZTEST).

But if I think a little deeper, I think there are two primary approaches:

1. ZTEST + FFF tests built into the firmware. Python does minimal serial IO to start the test and check the ZTEST output.
2. Shell commands in firmware, pytest is the main test controller. Code under test is activated via shell commands over serial from pytest.

I think there's merit to both approaches.

Option 1:
+ Has direct access to firmware functions and data. Better than option 2 where all control has to be routed through shell commands.
+ (minor) Does not require external python script to run tests, automatically run on boot
+ (minor) Same test frameworks used in all firmware tests (unit, integration, emulation, hardware)

Option 2:
+ pytest is a more powerful and flexible test framework

I think we should choose the option that seems like it will scale the best as we add more tests. I slightly prefer option 1, but I could be convinced that option 2 is better. 

What do you think?

samples/test/verify.py
```diff
@@ -0,0 +1,93 @@
...
+
+    # WiFi credentials are options (nRF9160dk does not need them)
+    if 'wifi/ssid' in settings:
+        set_setting(ser, 'wifi/ssid', settings['wifi/ssid'])
+
+    if 'wifi/psk' in settings:
+        set_setting(ser,'wifi/psk', settings['wifi/psk'])
+
+    set_setting(ser,'golioth/psk-id', settings['golioth/psk-id'])
+    set_setting(ser,'golioth/psk', settings['golioth/psk'])
```
mniestroj, line 40: Just minor comments about possible improvements:
1. I'd prefer `yaml` - just looks nicer IMO,
2. Would create something among those lines in `yaml`:
```
settings:
    golioth/psk-id: mypskid
    golioth/psk: mypsk
    wifi/ssid: myssid
    wifi/psk: mywifipsk
```
and load it by iterating through all `settings` node, so that we don't need explicitly list that in Python code.

mniestroj (approved):
It's basic test, but it should already give us feedback about reliability of the tests on nrf52840 or WiFi in general.

samples/test/verify.py
ncmiller: The main reason I went with JSON is that it is built-in to Python. But YAML is not a big deal, just need `pip install pyyaml`.

PR #246, opened 2022-08-06
net: golioth: enable Extended Master Secret support in mbedTLS

mniestroj:
Enable RFC 7627 [1] by default, which addresses some security problems.

[1] https://datatracker.ietf.org/doc/html/rfc7627

TODO:
- [ ] check if enabling it gives any value - we should enforce it to make it better


mniestroj:
Looks like mbedTLS does not support enforcing Extended Master Secret. If client received ServerHello without that extension, then handshake continues without using Extended Master Secret. This means that if both client and server treat that extension as optional, then there is no value in enabling it, as attacker (man-in-the-middle) can still compromise connection security.

See: https://datatracker.ietf.org/doc/html/rfc7627#section-4
> Clients and servers SHOULD NOT accept handshakes that do not use the
   extended master secret, especially if they rely on features like
   compound authentication that fall into the vulnerable cases described
   in [Section 6.1](https://datatracker.ietf.org/doc/html/rfc7627#section-6.1).

PR #247, opened 2022-08-08
golioth_settings: add int64 support

ncmiller:
The backend now supports CBOR encoding of integers.
Add additional type for int64 settings.

CBOR-encoded integers only use as many bytes as
necessary, but we will treat them all as uint64_t
after decoding.

Note: The Golioth backend does not support
integers greater than INT64_MAX (9223372036854775807),
so there's no need to support a uint64 type.

### Testing

Tested with local backend and local UI on branches with CBOR-encoding fixes:

Backend: https://github.com/golioth/golioth/pull/168
UI: https://github.com/golioth/golioth-web/pull/49

### DIFF
--- a/include/net/golioth/settings.h
+++ b/include/net/golioth/settings.h
@@ -70,6 +70,7 @@ enum golioth_settings_value_type {
 	GOLIOTH_SETTINGS_VALUE_TYPE_BOOL,
 	GOLIOTH_SETTINGS_VALUE_TYPE_FLOAT,
 	GOLIOTH_SETTINGS_VALUE_TYPE_STRING,
+	GOLIOTH_SETTINGS_VALUE_TYPE_INT64,
 };
 
 /**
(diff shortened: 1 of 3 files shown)


include/net/golioth/settings.h
```diff
@@ -70,6 +70,7 @@ enum golioth_settings_value_type {
 	GOLIOTH_SETTINGS_VALUE_TYPE_BOOL,
 	GOLIOTH_SETTINGS_VALUE_TYPE_FLOAT,
 	GOLIOTH_SETTINGS_VALUE_TYPE_STRING,
+	GOLIOTH_SETTINGS_VALUE_TYPE_INT64,
```
mniestroj, line 73: Commit message says:
> CBOR-encoded integers only use as many bytes as
necessary, but we will treat them all as uint64_t
after decoding.

It should be `s/uint64_t/int64_t/`, right?

mniestroj (approved):

mniestroj (approved):

PR #248, opened 2022-08-08
RPC and Settings Build Fixes

ncmiller:
There were build failures in samples when both RPC and Settings were enabled, so these commits fix those issues.

### DIFF
--- a/net/golioth/Kconfig
+++ b/net/golioth/Kconfig
@@ -41,6 +41,7 @@ config GOLIOTH_FW
 
 config GOLIOTH_RPC
 	bool "Remote Procedure Call"
+	select QCBOR
 	help
 	  Enable Golioth Remote Procedure Call feature.
 
@@ -54,6 +55,7 @@ config GOLIOTH_RPC_MAX_NUM_METHODS
 
 config GOLIOTH_SETTINGS
 	bool "Settings cloud service"
+	select QCBOR
 	help
 	  Enable Golioth Settings feature. Not to be confused
 	  with Zephyr's own Settings subsystem.
--- a/net/golioth/rpc.c
+++ b/net/golioth/rpc.c
@@ -218,9 +218,9 @@ static int golioth_rpc_observe(struct golioth_client *client)
 	return golioth_send_coap(client, &packet);
 }
 
-void on_message(struct golioth_client *client,
-		struct coap_packet *rx,
-		void *user_arg)
+static void on_message(struct golioth_client *client,
+		       struct coap_packet *rx,
+		       void *user_arg)
 {
 	k_mutex_lock(&client->rpc_mutex, K_FOREVER);
 	coap_response_received(rx, NULL, &client->rpc.observe_reply, 1);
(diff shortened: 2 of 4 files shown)


net/golioth/Kconfig
```diff
@@ -41,6 +41,8 @@ config GOLIOTH_FW
 
 config GOLIOTH_RPC
 	bool "Remote Procedure Call"
+	default y
```
mniestroj, line 44: I like all the fixes and improvements **besides** enabling GOLIOTH_RPC and GOLIOTH_SETTINGS by default. IMO those should be enabled per application need, same as GOLIOTH_SYSTEM_CLIENT and GOLIOTH_FW.

mniestroj (changes requested):

net/golioth/Kconfig
ncmiller: That's fine with me. I'll remove the `default y` lines.

net/golioth/Kconfig
```diff
@@ -41,6 +41,7 @@ config GOLIOTH_FW
 
 config GOLIOTH_RPC
 	bool "Remote Procedure Call"
+	depends on QCBOR
```
mniestroj, line 44: Looking once again into that... `select QCBOR` would probably be better. This is what we use with `GOLIOTH_FW` and `LOG_BACKEND_GOLIOTH`.

ncmiller: Yeah, I chose to use `depends on` based on a quick read of the recommendations here: https://docs.zephyrproject.org/latest/build/kconfig/tips.html?highlight=kconfig%20tips#select-recommendations

> Avoid selecting symbols with prompts or dependencies. Prefer depends on. If depends on causes annoying bloat in configuration files, consider adding a Kconfig default for the most common value.

> Rare exceptions might include cases where you’re sure that the dependencies of the selecting and selected symbol will never drift out of sync, e.g. when dealing with two simple symbols defined close to one another within the same if.

> Common sense applies, but be aware that select often causes issues in practice. depends on is usually a cleaner and safer solution.

ncmiller: Perhaps `select` is better in this case, since QCBOR doesn't have any dependencies?

ncmiller: I'll go ahead and change to `select`, for consistency.

mniestroj: >Perhaps select is better in this case, since QCBOR doesn't have any dependencies?

The dependency problem is a real thing. In Buildroot for example, which also is Kconfig based, if you select an option with dependencies (`depends on` entries), then all of those dependencies need to be duplicated (all `depends on` entries need to be copy-pasted) for the introduced option that does `select`. I haven't seen such practice with Zephyr, though. And that is not very developer-friendly.

> I'll go ahead and change to select, for consistency.

We can change `select` to `depends on` all at once if `select` will raise any issues. So yeah, consistency is the primary goal with my request here.

mniestroj (approved):

PR #249, opened 2022-08-09
settings_shell: add list command

ncmiller:
Added a command to list all settings.

Also simplified the shell help strings to fit on one line,
since the multi-line strings were not being formatted properly
at runtime (indentation not displayed correctly).

Note: CONFIG_SETTINGS_SHELL is similar (it also provides a
settings list command), but we can't use it because:

1. The name will clash (both shells have name "settings").
2. We need to maintain compatibility with external
   goliothctl tool, which relies on the --json option and
   is not available in the Zephyr settings shell.

### DIFF
--- a/samples/common/settings_shell.c
+++ b/samples/common/settings_shell.c
@@ -20,6 +20,10 @@ struct settings_read_callback_params {
 	bool json_output;
 };
 
+struct settings_list_callback_params {
+	const struct shell *shell_ptr;
+};
+
 static int cmd_settings_set(const struct shell *shell, size_t argc, char *argv[])
 {
 	bool json_output;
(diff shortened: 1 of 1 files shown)


samples/common/settings_shell.c
```diff
@@ -180,23 +184,46 @@ static int cmd_settings_get(const struct shell *shell, size_t argc,
...
-		"usage:\n"
-		"$ settings get <key>\n"
-		"example:\n"
-		"$ settings get golioth/psk\n"
-		"$ settings get golioth/psk-id",
+		"get a setting (usage: settings get <key> [--json])",
 		cmd_settings_get, 2, 1),
+	SHELL_CMD_ARG(list, NULL,
+		"list all settings (usage: settings list)",
+		cmd_settings_list, 1, 1),
```
mniestroj, line 226:
```suggestion
		cmd_settings_list, 1, 0),
```
Second number represents number of optional values I think. We don't have such with `list` command, right?

ncmiller: Good catch, it should be 0. 

I had originally planned on supporting a `--json` option, but realized it's not really feasible to create a full JSON listing (e.g. array) because the callback is passed individual settings.

mniestroj (approved):

PR #250, opened 2022-08-10
settings: option to persist Golioth settings

ncmiller:
### Main commit

Settings will be stored to persistent storage (via Zephyr's
Settings subsystem), but only if the user's settings
callback returns success, indicating the setting is
known and valid.

The intended use is for users that
need access to their settings before connecting to
Golioth. It's up to the user to retrieve the settings
out of the Zephyr Settings subsystem, if desired.

### Additional related commit

settings_shell: support to get nonprintable values 

If the setting is not a printable string, then "settings get"
will not display anything useful.

If the value contains nonprintable characters, we will
hex dump the value.

### DIFF
--- a/net/golioth/settings.c
+++ b/net/golioth/settings.c
@@ -10,6 +10,7 @@
 #include <qcbor/posix_error_map.h>
 #include <qcbor/qcbor.h>
 #include <qcbor/qcbor_spiffy_decode.h>
+#include <settings/settings.h>
 #include <stdio.h>
 #include <assert.h>
 
@@ -93,6 +94,27 @@ static int send_status_report(struct golioth_client *client,
 	LOG_HEXDUMP_DBG(response_buf, nbytes, "Response");
 
 	return send_coap_response(client, response_buf, nbytes);
+}
+
+static void save_setting(const char *key, const struct golioth_settings_value *value)
+{
+	int err;
+	char name[64] = {};
+
+	/*
+	 * Use the "gsettings" package in the Zephyr settings subsystem
+	 * to store all settings.
+	 */
+	snprintf(name, sizeof(name), "gsettings/%s", key);
+
+	if (IS_ENABLED(CONFIG_GOLIOTH_SETTINGS_PERSIST)) {
+		LOG_INF("Saving setting: %s", name);
+		err = settings_save_one(name, value, sizeof(struct golioth_settings_value));
+		if (err) {
+			LOG_ERR("Failed to save setting: %d", err);
+			return;
+		}
+	}
 }
 
 static int on_setting(const struct coap_packet *response,
@@ -192,6 +214,8 @@ static int on_setting(const struct coap_packet *response,
 
 			if (setting_status != GOLIOTH_SETTINGS_SUCCESS) {
 				cumulative_status = setting_status;
+			} else {
+				save_setting(key, &value);
 			}
 		}
 
--- a/include/net/golioth/settings.h
+++ b/include/net/golioth/settings.h
@@ -32,6 +32,14 @@
  * 4. For each setting, this library calls the user-registered callback.
  * 5. This library reports status of applying settings to cloud.
  *
+ * If the user's callback returns GOLIOTH_SETTINGS_SUCCESS, then the
+ * setting is automatically stored to persistent flash via the Zephyr
+ * Settings subsystem if CONFIG_GOLIOTH_SETTINGS_PERSIST is 'y'.
+ * Storing the setting in flash can be useful for applications that
+ * need the setting before the device connects to Golioth.
+ * For this use-case, it's the user's responsibility to retrieve
+ * the setting from the Zephyr Settings subsystem, if desired.
+ *
  * This library is responsible for interfacing with the cloud
  * and has no knowledge of the specific settings.
  * @{
--- a/net/golioth/Kconfig
+++ b/net/golioth/Kconfig
@@ -104,6 +104,14 @@ config GOLIOTH_SETTINGS
 	  Enable Golioth Settings feature. Not to be confused
 	  with Zephyr's own Settings subsystem.
 
+config GOLIOTH_SETTINGS_PERSIST
+	bool "Save Golioth settings to persistent storage"
+	depends on GOLIOTH_SETTINGS
+	default y
+	help
+	  Automatically save settings received from Golioth cloud to persistent
+	  storage via the Zephyr Settings subsystem.
+
 config GOLIOTH_FW_PACKAGE_NAME_MAX_LEN
 	int "Maximum length of package name"
 	default 32
--- a/samples/common/settings_shell.c
+++ b/samples/common/settings_shell.c
@@ -101,6 +101,7 @@ static int settings_read_callback(const char *key,
 	ssize_t num_read_bytes = MIN(len, SETTINGS_MAX_VAL_LEN);
 	uint8_t buffer[num_read_bytes + 1];
 	struct settings_read_callback_params *params = param;
+	bool value_is_printable = true;
 
 	/* Process only the exact match and ignore descendants of the searched name */
 	if (settings_name_next(key, NULL) != 0) {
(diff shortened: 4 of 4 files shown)


net/golioth/settings.c
```diff
@@ -192,6 +214,8 @@ static int on_setting(const struct coap_packet *response,
 
 			if (setting_status != GOLIOTH_SETTINGS_SUCCESS) {
 				cumulative_status = setting_status;
+			} else {
+				save_setting(key, &value);
```
mniestroj, line 218: I would change this to:
```
if (setting_status == GOLIOTH_SETTINGS_SUCCESS) {
	save_setting(key, &value);
} else {
	cumulative_status = setting_status;	
}
```
so that our mind doesn't have to negate twice (`!=` followed by `else`). It should improve readability.

net/golioth/settings.c
```diff
@@ -95,6 +96,27 @@ static int send_status_report(struct golioth_client *client,
...
+
+	/*
+	 * Use the "gsettings" package in the Zephyr settings subsystem
+	 * to store all settings.
+	 */
+	snprintf(name, sizeof(name), "gsettings/%s", key);
+
+	if (IS_ENABLED(CONFIG_GOLIOTH_SETTINGS_PERSIST)) {
+		LOG_INF("Saving setting: %s", name);
+		err = settings_save_one(name, value, sizeof(struct golioth_settings_value));
```
mniestroj, line 112: Nit: I prefer to use `sizeof(*value)`, so if the type changes in function argument, we don't have to update function body. It is also easier to verify that proper size is used (it is all part of single line, no need to look into the whole function to check if `value` is really a pointer to `struct golioth_settings_value`).

What do you think?

net/golioth/settings.c
```diff
@@ -95,6 +96,27 @@ static int send_status_report(struct golioth_client *client,
 	return send_coap_response(client, response_buf, nbytes);
 }
 
+static void save_setting(const char *key, const struct golioth_settings_value *value)
+{
+	int err;
+	char name[64] = {};
```
mniestroj, line 102: Shouldn't we use something like `sizeof("gsettings/") + GOLIOTH_SETTINGS_MAX_NAME_LEN` instead of `64`?

net/golioth/settings.c
```diff
@@ -95,6 +96,27 @@ static int send_status_report(struct golioth_client *client,
...
+
+	/*
+	 * Use the "gsettings" package in the Zephyr settings subsystem
+	 * to store all settings.
+	 */
+	snprintf(name, sizeof(name), "gsettings/%s", key);
```
mniestroj, line 108: `snprintf` doesn't guarantee to append `\0` at the end, if `key` will be large enough. I guess we should check return value from `snprintf` to make sure we fit into the `name` buffer.

From `man snprintf`:
> The  functions  snprintf() and vsnprintf() do not write more than size bytes (including the terminating null byte ('\0')).  If the output was truncated due to this limit, then the return value is the number of characters (excluding the terminating null byte) which would have been  written  to the  final  string  if enough space had been available.  Thus, a return value of size or more means that the output was truncated.

net/golioth/settings.c
mniestroj, line 112: I don't think saving:
```
struct golioth_settings_value {
	enum golioth_settings_value_type type;
	union {
		bool b;
		float f;
		struct {
			const char *ptr; /* not NULL-terminated */
			size_t len;
		} string;
		int64_t i64;
	};
};
```
and restoring it after firmware upgrade (changing firmware build in general) is 100% safe, because of the following reasons:
1. `struct golioth_settings_value` might change over time - we might further extend it, which will make it binary incompatible,
2. there are endianess problems - potentially user might change endianness of between builds for targets that support both little and big endian (very unlikely, but still),
3. I don't know C standard enough to be sure that `enum` will remain the same size even if we add more values to it, `bool` will occupy the same amount of bytes at the same offset, etc.; all those are subject to toolchain upgrade, so if something is undefined by standard, it is potentially a bug to rely on that.

In the end, to be 100% sure that we save something persistently and then can restore it from by different firmware build, then we need to either:
1. be very strict on how we write those settings, e.g. using `__packed` and paying attention to endianness,
2. use a serialization library like CBOR to handle all the complexities of back and forth compatibility between firmware builds.

In the past I have run into the same kind of problem and figured that it is much better to use CBOR for serializing typed settings. This includes the complexity of writing all the code to serialize and deseriallize CBOR, but it also solves all the complexities and corner cases of how binary representation of the same structure will look like (which is why the serialization libraries exist in the first place). CBOR also allows to save `bool` into single byte, whereas with current code it would occupy 8 bytes (in most platforms/cases).

mniestroj: Besides, storing string the way it is now is not going to work (just pointer and length are stored, not the real payload).

net/golioth/settings.c
ncmiller: Sure, I can pass `sizeof(name) - 1` to ensure the last byte is NULL.

I was using this as a reference, but I suppose the C++ standard deviates from C in this regard: https://cplusplus.com/reference/cstdio/snprintf/

> If the resulting string would be longer than n-1 characters, the remaining characters are discarded and not stored, but counted for the value returned by the function.

> A terminating null character is automatically appended after the content written.

net/golioth/settings.c
mniestroj: Well, CBOR also adds type information, so you can easily change the type from `int64_t` to `float` between FW releases and when loading settings from persistent storage one can detect that previously it was an `int64_t` and now it is `float`.

>and I think endianness is not a concern, since it's extremely unlikely a user will change endianness between FW releases

Unlikely, but possible :)

IMO if we are going to add the type of variable to stored settings, then added complexity isn't much different from including CBOR as a dependency. But... hard to say without that being implemented. I guess we would need a sample that actually read those settings, as without it users might wonder how to get into those settings (we are getting way beyond original idea of "dump the structure, read the structure" which was relatively straightforward for users).

net/golioth/settings.c
mniestroj: Still would be better to log an error/warning or even not attempt to write setting with truncated name.

net/golioth/settings.c
ncmiller: > Well, CBOR also adds type information, so you can easily change the type from int64_t to float between FW releases and when loading settings from persistent storage one can detect that previously it was an int64_t and now it is float.

True. Although I tend to think that's the user's problem to solve.

> Unlikely, but possible

You're right, but I also think that making the interface more complicated for the .1% hypothetical user is not necessarily a good tradeoff to make.

> IMO if we are going to add the type of variable to stored settings, then added complexity isn't much different from including CBOR as a dependency. But... hard to say without that being implemented. I guess we would need a sample that actually read those settings, as without it users might wonder how to get into those settings (we are getting way beyond original idea of "dump the structure, read the structure" which was relatively straightforward for users)

With all of the points you've made in the PR, I think it's probably not worth trying to store the settings in flash at all. There are just too many things that can go wrong, and the value added is not worth it. The user can save settings however they want (and if they want), and we shouldn't jump through a bunch of hoops trying to do things automatically for them, because it's just too awkward and too much code.

So, I think I'm just going to close this PR.

FWIW, in the ESP-IDF SDK we store the settings to flash automatically (the NVS library has functions to set/get int, bool, float, string, so we are just calling those functions), and there's no regard for possible endianness or type changes between FW releases (i.e. we assume the user will solve this problem if it comes up). I believe the type information is stored in flash by the NVS library, but I'd have to check. I'll probably leave that SDK as is, since I don't think it's terribly important for the SDKs to be consistent on this point.

mniestroj: >FWIW, in the ESP-IDF SDK we store the settings to flash automatically (the NVS library has functions to set/get int, bool, float, string, so we are just calling those functions), and there's no regard for possible endianness or type changes between FW releases (i.e. we assume the user will solve this problem if it comes up). I believe the type information is stored in flash by the NVS library, but I'd have to check. I'll probably leave that SDK as is, since I don't think it's terribly important for the SDKs to be consistent on this point.

It would be good if Zephyr supported such feature, so we could just utilize that, same as with ESP-IDF. Storing settings without giving users at least an example how to restore them is like half-baked solution IMO, thus not sure about the overall value this gives to users. And overall the place for this feature is rather Zephyr upstream, because storing settings is orthogonal task to cloud connection. There are also many non-volatile storage options with Zephyr, like EEPROM, NVS, Settings, FCB, FS, raw flash access, etc., so it is hard to satisfy majority of users with something that doesn't turn out to be a subsystem ("typed settings").

ncmiller: Agreed. I think if Zephyr were to support typed settings, we could probably resurrect this PR. For now, I think I'll just close it.

PR #252, opened 2022-08-12
net/golioth: remove hexdump of tx packet

ncmiller:
If CONFIG_LOG_GOLIOTH_LOG_LEVEL_DBG is set to 'y'
then the system will freeze at boot after a couple
of seconds (tested on nRF9160dk), which I suspect
is due to the hexdump of tx packets, because when
the hexdump is removed, the system boots normally.

### DIFF
--- a/net/golioth/golioth.c
+++ b/net/golioth/golioth.c
@@ -346,8 +346,6 @@ int golioth_send_coap(struct golioth_client *client, struct coap_packet *packet)
 {
 	int err;
 
-	LOG_HEXDUMP_DBG(packet->data, packet->offset, "TX CoAP");
-
 	err = golioth_send(client, packet->data, packet->offset, 0);
 	if (err) {
 		return err;


net/golioth/golioth.c
```diff
@@ -346,8 +346,6 @@ int golioth_send_coap(struct golioth_client *client, struct coap_packet *packet)
 {
 	int err;
 
-	LOG_HEXDUMP_DBG(packet->data, packet->offset, "TX CoAP");
-
```
mniestroj, line 350: There is no `CONFIG_LOG_GOLIOTH_LOG_LEVEL_DBG`, which commit message references.

Commit message does not mention it, but the reason why log message will result in freeze is that for each generated log message, we will once again enter the same code path when trying to send such log message to Golioth. This means that another log message will be generated, and then again, again and again. So overall infinite log messages will be generated, which creates kind of deadlock.

There are potentially more log messages that could create such issue, also in the receiving path (for each sent log message cloud will respond with CoAP packet, so that could once again generate log message, creating similar infinite loop, but with cloud in the middle).

mniestroj (approved):
Definitely an improvement, but we need to get a closer look at some point for a more generic solution to this (like observing which thread generated log message - if logging thread, then just drop early in logging backend implementation).

net/golioth/golioth.c
ncmiller: > There is no CONFIG_LOG_GOLIOTH_LOG_LEVEL_DBG, which commit message references.

I meant `CONFIG_GOLIOTH_LOG_LEVEL_DBG`. I'll change the commit message.

I'll also add the reason to the commit message. Thanks for explaining, I suspected that might have been what was happening, but I wasn't sure.

mniestroj (approved):

PR #253, opened 2022-08-12
settings: send response as CBOR

ncmiller:
* settings: send response as CBOR
* samples/settings: LOOP_DELAY_S is int64 instead of float 

---

~(Aug 16, 2022) Leaving this as a Draft PR for now, until support on the backend is deployed to production, at which point I'll turn it into a proper PR.~

(Aug 25, 2022) Backend support for CBOR types and `.c/status` is deployed, so marking PR as "Ready for review"

### DIFF
--- a/net/golioth/settings.c
+++ b/net/golioth/settings.c
@@ -39,8 +39,8 @@ LOG_MODULE_DECLARE(golioth);
 
 #define GOLIOTH_SETTINGS_PATH ".c"
 #define GOLIOTH_SETTINGS_STATUS_PATH ".c/status"
-#define GOLIOTH_SETTINGS_MAX_NAME_LEN 63
-#define GOLIOTH_SETTINGS_MAX_RESPONSE_LEN 63
+#define GOLIOTH_SETTINGS_MAX_NAME_LEN 63 /* not including NULL */
+#define GOLIOTH_SETTINGS_MAX_RESPONSE_LEN 64
 
 static int send_coap_response(struct golioth_client *client,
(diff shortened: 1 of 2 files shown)


net/golioth/settings.c
```diff
@@ -66,7 +66,7 @@ static int send_coap_response(struct golioth_client *client,
 
 	err = coap_append_option_int(&packet,
 				     COAP_OPTION_CONTENT_FORMAT,
-				     COAP_CONTENT_FORMAT_APP_JSON);
+				     COAP_CONTENT_FORMAT_APP_CBOR);
```
mniestroj, line 69: Just a reminder to squash that before final version.

net/golioth/settings.c
```diff
@@ -93,6 +94,28 @@ static int send_status_report(struct golioth_client *client,
 	LOG_HEXDUMP_DBG(response_buf, nbytes, "Response");
 
 	return send_coap_response(client, response_buf, nbytes);
+#endif
+
+	uint8_t response_buf[GOLIOTH_SETTINGS_MAX_RESPONSE_LEN + 1] = {};
```
mniestroj, line 99: I don't think there is a reason for `+ 1` in this context. We are creating binary data (CBOR), so initializing also seems unnecessary.

ncmiller:
Ignoring the checkpatch failure `Avoid CamelCase`, since this comes from  CBOR and there's nothing we can do about it.

mniestroj (approved):

mniestroj (approved):

PR #262, opened 2022-08-26
golioth: lightdb: split out as new module

mniestroj:
Move lightdb related functions out of golioth.c module, in order to get
better code separation and readability. This will also make more space for
low-level packet handling in golioth.c module.


net/golioth/lightdb.c
```diff
@@ -0,0 +1,167 @@
...
+ */
+
+#include <net/golioth.h>
+
+#include "coap_utils.h"
+
+#include <zephyr/logging/log.h>
+LOG_MODULE_REGISTER(lightdb, CONFIG_GOLIOTH_LOG_LEVEL);
+
+int golioth_lightdb_get(struct golioth_client *client, const uint8_t *path,
```
ncmiller, line 14: Does it make sense to declare these functions in a new file `lightdb.h`? Currently, they are defined in `golioth.h`.

mniestroj: Makes sense, it will be more consistent with rpc and settings services. Also it seems easier to find lightdb related APIs, as they are group in one file.

Updated!

ncmiller (approved):

PR #269, opened 2022-09-01
[CoAP rework] move lightdb, lightdb_stream, rpc and settings to 'coap_req'

mniestroj:
Closes: #269


mniestroj:
@ncmiller @szczys @ChrisGammell This contains new high-level API proposal for all the services (besides logging). All samples updated. Any comments are welcome. The underneath mechanism (coap_req) still needs some love and better testing/error handling. But it will mostly stay as is.

samples/lightdb/get/src/main.c
```diff
@@ -9,14 +9,10 @@ LOG_MODULE_REGISTER(golioth_lightdb, LOG_LEVEL_DBG);
 
 #include <net/golioth/system_client.h>
 #include <samples/common/wifi.h>
-#include <zephyr/net/coap.h>
```
ncmiller, line 12: Should update the copyright at the top of the file to `2021-2022`.

samples/lightdb/get/src/main.c
```diff
@@ -25,113 +21,71 @@ static void golioth_on_connect(struct golioth_client *client)
...
-			/*
-			 * Clear the reply so it can be used again in the
-			 * future. This is effectively enforcing a 5 second
-			 * timeout on requests.
-			 */
-			coap_reply_clear(reply);
-
-			k_mutex_unlock(&coap_reply_mutex);
-		}
+		k_sleep(K_SECONDS(5));
```
ncmiller, line 89: This is a much cleaner sample now, nice work!

net/golioth/lightdb.c
```diff
@@ -6,49 +6,108 @@
...
+	 */
+	if (!total) {
+		total = rsp->off + rsp->len;
 	}
 
-	err = coap_packet_append_uri_path_from_stringz(&packet, path);
-	if (err) {
-		LOG_ERR("Unable add uri path to packet");
-		return err;
+	/* Reallocate buffer if we need more space */
```
ncmiller, line 76: Comment doesn't match the code.

include/net/golioth/lightdb.h
```diff
@@ -158,6 +158,46 @@ int golioth_lightdb_delete_cb(struct golioth_client *client, const uint8_t *path
...
+ * Asynchronously push new value to LightDB Stream and let @p cb be invoked when server
+ * acknowledges it or some error condition happens.
+ *
+ * @param client Client instance
+ * @param path LightDB Stream resource path
+ * @param format Format of payload
+ * @param data Payload data
+ * @param data_len Payload length
+ * @param[in] cb Callback executed on response received, timeout or error
+ * @param[in] user_data User data passed to @p cb
```
ncmiller, line 170: Inconsistent usage of param in/out annotations.

include/net/golioth/lightdb.h
```diff
@@ -158,6 +158,46 @@ int golioth_lightdb_delete_cb(struct golioth_client *client, const uint8_t *path
...
+ *
+ * @retval 0 On success
+ * @retval <0 On failure
+ */
+int golioth_stream_push_cb(struct golioth_client *client, const uint8_t *path,
```
ncmiller, line 175: Nice decision to use the word "push", very descriptive.

ncmiller (commented):
Looking good so far

include/net/golioth/lightdb.h
mniestroj: Right, will fix that. While we are on that topic, what should be the in/out annotation for `struct golioth_client *client` argument? `@param[in]` or `@param[in,out]`?

include/net/golioth/lightdb.h
mniestroj: Was struggling with the name for this function. I decided to drop `lightdb` component from it. This is something that `goliothctl` does as well, with `goliothctl stream` subcommand. Additionally it makes API calls much smaller.

I guess we should move those functions to `stream.c` and `net/golioth/stream.h` files, right?

net/golioth/lightdb.c
mniestroj: Fixed

samples/lightdb/get/src/main.c
mniestroj: Updated.

include/net/golioth/lightdb.h
mniestroj: Updated annotations in `lightdb.h`.

include/net/golioth/lightdb.h
mniestroj: Unfortunately we are not 100% consistent. IMO one of the problems causing that is that one service name "LightDB Stream" has another service "LightDB" in its name. That is confusing. People tend to use short names (abbreviations), so sometimes "LightDB" might be used when refering to "LightDB Stream". Docs are all good now, but I mean informal use of it (like between team members, during a call, etc.). I heard at least few times a question like "by LightDB you mean LightDB State, right?".

So in theory we have those 3 most commonly used terminology that is around:

1) LightDB | LightDB Stream

This is most common and it is what https://docs.golioth.io uses. IMO the biggest downside of that is the use of "LightDB" in both. In my engineer's mind, it would make sense if the latter (LightDB Stream) would be a subset of the former (LightDB), but not when it is a totally different service. The origins of both are similar, but I think that the use case of both is so different, that names of those services do not reflect that enough.

2) LightDB State | LightDB Stream

Not sure this is used formally, but it is used sometimes. The advantage is that using one of those names is always clear which service we are talking about.

3) LightDB | Stream

Yeah, we use that as well. Here are examples:
- `goliothctl lightdb` and `goliothctl stream` commands
- https://docs.golioth.io/reference/rest-api/openapi/ uses `.../{deviceId}/data` (LightDB) and `.../{deviceId}/stream` (LightDB Stream) endpoints.

I like it mostly because it is short, single word. All API calls are shorter that way. Also, there is no confusion when talking about "LightDB", as there is only one service with that word in name.

ncmiller: We can stick with the names you are proposing. There's still not much consistency in docs and other places, but we won't be able to solve that problem right now. The names you've got make sense and are short.

> I guess we should move those functions to stream.c and net/golioth/stream.h files, right?

Yeah, I think I would prefer this, to make it easier to find them and to be consistent with the API names.

mniestroj: Created `stream.h` and `stream.c` and left `golioth_stream_push{,_cb}` APIs, as originally proposed.

include/net/golioth/lightdb.h
```diff
@@ -20,77 +21,139 @@
...
+ * @retval 0 On success
+ * @retval <0 On failure
+ */
+int golioth_lightdb_set_cb(struct golioth_client *client, const uint8_t *path,
+			   enum golioth_content_format format,
+			   const uint8_t *data, size_t data_len,
+			   golioth_req_cb_t cb, void *user_data);
+
+/**
+ * @brief Set value to Golioth's LightDB (synchronously)
```
ncmiller, line 88: Are you still planning to add `golioth_lightdb_set_auto` on top of `golioth_lightdb_set`, perhaps in a future PR?

ncmiller (approved):
Looks good to me. 

It seems that CI is not running since the base branch is not `main`. You could defer CI checks until after this merges to `golioth-dfu`, or we could update the workflows to run on any PR, regardless of base branch. Which do you think is best here?

include/net/golioth/lightdb.h
mniestroj: I've added #277. I was not working on that last days, it just stayed in a Proof of Concept state. It is something I would like to do in separate PR anyway. I think I can take a little time after this PR gets merged, to cleanup and get it out of "draft" stage.

mniestroj:
> It seems that CI is not running since the base branch is not main. You could defer CI checks until after this merges to golioth-dfu, or we could update the workflows to run on any PR, regardless of base branch. Which do you think is best here?

I think that building PRs to any branches make sense.

ncmiller:
> > It seems that CI is not running since the base branch is not main. You could defer CI checks until after this merges to golioth-dfu, or we could update the workflows to run on any PR, regardless of base branch. Which do you think is best here?
> 
> I think that building PRs to any branches make sense.

Handling that in https://github.com/golioth/golioth-zephyr-sdk/pull/279.

PR #270, opened 2022-09-02
[CoAP rework] coap_req: introduce new helper module

mniestroj:
Introduce new helper module, which will handle most low-level CoAP specific
tasks like:

 * contructing CoAP packets for the most common use cases,
 * keeping track of pending (unacked) CoAP packets,
 * retransmitting packets with exponential backoff,
 * requesting wakeup from poll() system call, when new packet is to be
   transmitted or there is a retransmission of pending packet needed,
 * handling block-wise transfers (requesting next blocks of data, if
   response contains information about block number),
 * providing thread-safe APIs.

Add `golioth_prepare_poll()` function will allow to get file descriptor and next
timeout value, so those values can be passed to subsequent poll() system call. 
This is used by `system_client.c` implementation, but the idea is to split it into 
application specific custom "system client" implementation.

TODO:
- [x] DFU samples fails to build for ESP32 (DRAM overflow)
- [x] to be tested with nRF91 modem sockets, where wakeup might not work (due to no suport of waking up from `poll()`)
- [x] wrap offloaded sockets to allow poll() be interrupted (part of #298)

### DIFF
--- a/net/golioth/coap_req.c
+++ b/net/golioth/coap_req.c
@@ -1,0 +1,635 @@
+/*
+ * Copyright (c) 2022 Golioth, Inc.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <zephyr/logging/log.h>
+LOG_MODULE_DECLARE(golioth);
+
+#include <stdlib.h>
+
+#include <zephyr/random/rand32.h>
+
+#include "coap_req.h"
+#include "coap_utils.h"
+#include "golioth_utils.h"
+
+#define COAP_RESPONSE_CODE_CLASS(code)	((code) >> 5)
+
+static int golioth_coap_req_send(struct golioth_coap_req *req)
+{
+	return golioth_send_coap(req->client, &req->request);
+}
+
+static void golioth_coap_pending_init(struct golioth_coap_pending *pending,
+				      uint8_t retries)
+{
+	pending->t0 = k_uptime_get_32();
+	pending->timeout = 0;
+	pending->retries = retries;
+}
+
+static void __golioth_coap_req_submit(struct golioth_coap_req *req)
+{
+	struct golioth_client *client = req->client;
+
+	sys_dlist_append(&client->coap_reqs, &req->node);
+}
+
+static void golioth_coap_req_submit(struct golioth_coap_req *req)
+{
+	struct golioth_client *client = req->client;
+
+	k_mutex_lock(&client->coap_reqs_lock, K_FOREVER);
+	__golioth_coap_req_submit(req);
+	k_mutex_unlock(&client->coap_reqs_lock);
+}
+
+static void golioth_coap_req_cancel(struct golioth_coap_req *req)
+{
+	sys_dlist_remove(&req->node);
+}
+
+static void golioth_coap_req_cancel_and_free(struct golioth_coap_req *req)
+{
+	LOG_DBG("cancel and free req %p data %p", req, req->request.data);
+
+	golioth_coap_req_cancel(req);
+	free(req->request.data);
+	free(req);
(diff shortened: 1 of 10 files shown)


net/golioth/coap_req.c
```diff
@@ -0,0 +1,709 @@
...
+
+			.get_next = NULL,
+			.get_next_data = NULL,
+
+			.user_data = req->user_data,
+		};
+
+		(void)req->cb(&rsp);
+
+		goto cancel_and_free;
```
ncmiller, line 300: In the blockwise case, after calling the callback, the return value is checked. Any reason we're not checking it here?

net/golioth/coap_req.c
```diff
@@ -0,0 +1,709 @@
...
+	err = golioth_coap_req_cb(client, method, pathv, format,
+				  data, data_len,
+				  golioth_req_sync_cb, &sync_data,
+				  flags);
+	if (err) {
+		LOG_WRN("Failed to make CoAP request: %d", err);
+		return err;
+	}
+
+	k_sem_take(&sync_data.sem, K_FOREVER);
```
ncmiller, line 586: Is it ever possible for `sync_data.sem` to never be given? Is it guaranteed to always be given?

ncmiller (approved):
I didn't do a super-detailed review, but I reviewed enough to get the general flow, and I got the impression that it's well-written and could be maintainable by someone else.

My main concerns would be possible memory leaks or unresponsiveness due to synchronization issues in error cases like dropped packets, lost network comms, etc. I don't see anything specific in the code that would indicate this is possible, but I also couldn't convince myself it's not possible (difficult due to sem take/give and alloc/free logic being split among different functions and branches), which could make it prone to regressions in the future (especially if the code is modified by someone else). Automated testing could help alleviate the concern (e.g. a memory leak test).

ncmiller:
> to be tested with nRF91 modem sockets, where wakeup might not work (due to no suport of waking up from poll())

Is this still a TODO?

net/golioth/coap_req.c
mniestroj: Actually only in case of blockwise transfer this return value is checked and if it is non-zero, then blockwise transfer is not continued. This is a special case and subject to refinement in the future. The idea was to use (just in blockwise transfer case) return code to signal whether application want to continue receiving next blocks of data or stop (e.g. when downloaded content does not fit into preallocated memory or in case of any other error).

This is part of asynchronous API discussion/improvement. For now this API is marked in Doxygen comments as "experimental", mainly because of the details on how to handle blockwise transfer. Also 2 issues were added to Jira, which described the direction of (possible) further improvements.

mniestroj:
> > to be tested with nRF91 modem sockets, where wakeup might not work (due to no suport of waking up from poll())
> 
> Is this still a TODO?

Tested. Unfortunately `poll()` does not support being interrupted with offloaded sockets (which nRF91 sockets are). Created https://golioth.atlassian.net/browse/DSDK-390 to address that.

net/golioth/coap_req.c
mniestroj: It is assumed so. However I realized that this is not true in case of connection close. In case of connection restart "system client" loop should continue resending / expiring pending packets with this implementation. This is bad (because packets from previous connection should not be "resent" with new connection), however it would unblock application thread at some point.

Added https://golioth.atlassian.net/browse/DSDK-390 to cover that.

ncmiller: As a defensive measure, you could add an assert after a long time, in case it can't be taken for any reason. Something like:

```c
assert(k_sem_take(&sync_data.sem, K_SECONDS(100)) == 0);
```

PR #271, opened 2022-09-02
Add mimxrt1060_evkb to all samples along with ethernet support

szczys:
* Add Ethernet to common samples to work with this, and all ethernet-based
boards.
* Add board-specific conf file
* Add include and check for golioth ethernet Kconfig symbol

This PR follows the example of samples/common/wifi, adding an Ethernet option to handle the needed DHCP operations for the connection type.


west-zephyr.yml
```diff
@@ -14,6 +14,8 @@ manifest:
           - net-tools
           - segger
           - tinycrypt
+          - hal_nxp
+          - lvgl
```
mniestroj, line 18: Please keep the list in alphabetic order. Why `lvgl`?

samples/common/include/samples/common/ethernet.h
```diff
@@ -0,0 +1,4 @@
+#ifndef __GOLIOTH_INCLUDE_GOLIOTH_ETH_H__
+#define __GOLIOTH_INCLUDE_GOLIOTH_ETH_H__
+void eth_connect(void);
+#endif
```
mniestroj, line 4:
```suggestion
#define __GOLIOTH_INCLUDE_GOLIOTH_ETH_H__

void eth_connect(void);

#endif /* __GOLIOTH_INCLUDE_GOLIOTH_ETH_H__ */
```

samples/common/ethernet.c
```diff
@@ -0,0 +1,19 @@
...
+#include <samples/common/ethernet.h>
+#include <zephyr/net/net_if.h>
+
+void eth_connect(void)
+{
+	LOG_INF("Connecting to Ethernet");
+	struct net_if *iface;
+	iface = net_if_get_default();
+	net_dhcpv4_start(iface);
+}
```
mniestroj, line 19: I don't think "Connecting to Ethernet" is a correct form. It is just starting DHCP. How about just putting `net_dhcpv4_start(net_if_get_default());` in samples instead of `eth_connect()`?.

samples/dfu/src/main.c
```diff
@@ -243,6 +244,9 @@ void main(void)
 		}
 	}
 
+	if (IS_ENABLED(CONFIG_GOLIOTH_SAMPLE_ETHERNET)) {
+		eth_connect();
+	}
 	if (IS_ENABLED(CONFIG_GOLIOTH_SAMPLE_WIFI)) {
 		LOG_INF("Connecting to WiFi");
 		wifi_connect();
```
mniestroj, line 252:
```suggestion
	if (IS_ENABLED(CONFIG_GOLIOTH_SAMPLE_WIFI)) {
		LOG_INF("Connecting to WiFi");
		wifi_connect();
	}
	
	if (IS_ENABLED(CONFIG_GOLIOTH_SAMPLE_DHCP_START)) {
		net_dhcpv4_start(net_if_get_default());
```

If we put that after "WiFi connect" then in the future we might handle the case when DHCP needs to be started for WiFi network. This might be needed for more "natve" WiFi support, like ESP32 (Espressif is working on adjusting to Zephyr WiFi APIs) or nRF WiFi.

samples/common/Kconfig
```diff
@@ -59,4 +59,12 @@ config GOLIOTH_SAMPLE_WIFI_PSK
 
 endif # (GOLIOTH_SAMPLE_WIFI && !GOLIOTH_SAMPLE_WIFI_SETTINGS) || WIFI_ESP32
 
+config GOLIOTH_SAMPLE_ETHERNET
+	bool "Ethernet DHCP for samples"
+	depends on NET_L2_ETHERNET
+	depends on NET_MGMT
+	depends on NET_DHCPV4
+	help
+	  Aquire an IP address from the DHCP server when using Ethernet
```
mniestroj, line 68:
```suggestion
	  Aquire an IP address from the DHCP server when using Ethernet.
```

samples/common/CMakeLists.txt
```diff
@@ -1,5 +1,6 @@
 zephyr_library_sources_ifdef(CONFIG_GOLIOTH_SAMPLE_SETTINGS_AUTOLOAD settings_autoload.c)
 zephyr_library_sources_ifdef(CONFIG_GOLIOTH_SAMPLE_SETTINGS_SHELL settings_shell.c)
 zephyr_library_sources_ifdef(CONFIG_GOLIOTH_SAMPLE_WIFI wifi.c)
+zephyr_library_sources_ifdef(CONFIG_GOLIOTH_SAMPLE_ETHERNET ethernet.c)
```
mniestroj, line 4: Alphabetic order, please.

west-zephyr.yml
szczys: Previously the default config for this board threw errors when lvgl wasn't present:

```
warning: LV_Z_HOR_RES_MAX (defined at boards/arm/mimxrt1060_evk/Kconfig.defconfig:65) defined without a type                                                                  
                                                                                                                                                                              
warning: LV_Z_VER_RES_MAX (defined at boards/arm/mimxrt1060_evk/Kconfig.defconfig:68) defined without a type                                                                  
Parsing /home/mike/golioth-zephyr-workspace/zephyr/Kconfig                                                                                                                    
Loaded configuration '/home/mike/golioth-zephyr-workspace/zephyr/boards/arm/mimxrt1060_evk/mimxrt1060_evk_defconfig'                                                          
Merged configuration '/home/mike/golioth-zephyr-workspace/zephyr/samples/basic/minimal/prj.conf'                                                                              
                                                                                                                                                                              
error: Aborting due to Kconfig warnings  
```

But it appears to be fixed now that our snapshot includes this commit so I'll remove this dependency: https://github.com/zephyrproject-rtos/zephyr/commit/8eead500717e733abab6ac3a0952cd60bea5f8ac

samples/common/include/samples/common/ethernet.h
szczys: Removing this file per your other advice.

samples/common/ethernet.c
szczys: Good call, I'm removing this file and following your suggested pattern.

samples/common/CMakeLists.txt
szczys: This is no longer needed.

szczys:
After the changes this builds for the mimxrt1060_evkb, but not for other configurations that don't have `CONFIG_NET_DHCPV4` selected. I'll be a little more clever and push some changes when I figure this out.

mniestroj:
I guess you should include `zephyr/net/dhcpv4.h` to fix this build issue.

mniestroj:
@szczys Nit: for consistency please use lowercase `samples` in commit messages, so that grepping git history with something like `git log --grep=samples:` would still be possible (with your commits included).

szczys:
I added https://github.com/golioth/golioth-zephyr-sdk/pull/271/commits/44a567106135972c070af681f8da3bd8b6b084ce to this PR to change the way we call network helper functions from samples:

- `net_config()` is run by all samples but will be empty unless other symbols like `CONFIG_GOLIOTH_SAMPLE_WIFI` are selected
- This is not a breaking change, existing apps that call `wifi_connect()` are still able to do so
- Future additions to `net_config()` can be made without changing the samples (new symbols can be added in the \<board\>.conf)

One question I have is whether I actually need to make the `CONFIG_GOLIOTH_SAMPLE_DHCP_START` dependent on `NET_L2_ETHERNET`. I need it for this board (and other Ethernet connections) but I don't know if there's a future use case for DHCP when ethernet will not be selected.

ncmiller (approved):
Looks good to me, thanks. I tested the `hello` and `test` samples on my RT1060 dev board on this branch, and they both worked.

PR #274, opened 2022-09-09
samples: lightdb_stream: add twister runtime test with pytest

mniestroj:
Add basic pytest runtime test, which runs:

`
  $ goliothctl $GOLIOTHCTL_OPTS stream listen $GOLIOTH_DEVICE_NAME
`

and verifies if temperature measurements are received. Right now an arbitrary number of 10 events is expected, before considering test as passed.

TODO:
- [ ] document which environment variables should be set before running this by twister with runtime test (not just build time)
- [ ] figure out if there is a better way of passing credentials than via environment variables fetched from Kconfig
- [ ] add runtime tests to CI, so that this code actually gets used


mniestroj:
Covered as part of #287 and #294

PR #275, opened 2022-09-09
Enable the nrf9160 CI workflow

szczys:


szczys:
This is  tested and both the mimxrt1060_evkb and nrf9160dk boards with CI. However, the nrf9160 is inconsistent on how long it takes to connect. We will not be adding that board to CI for now so I'm closing this draft PR.

PR #276, opened 2022-09-09
.github: Enable test_nrf9160dk.yml workflow

ncmiller:
Previously, the workflow was disabled due to unreliable cellular connection at bootup, which could result in the connection test failing.

Another self-hosted runner (mikes_nuc) was recently set up with a nrf9160dk which should have a more stable cellular connection, so the workflow can be re-enabled.


ncmiller:
Closing. Duplicate of https://github.com/golioth/golioth-zephyr-sdk/pull/275.

PR #281, opened 2022-09-15
samples: reduce k_malloc() heap size for ESP32

mniestroj:
ESP32 platform has just 176.5 kB RAM for application use. This includes 128 kB reserved for RAM, which leaves just 48.5 kB for non-heap (k_malloc) use.

Reduce size of k_malloc() heap to 96 kB, which is what zephyr/samples/board/esp32/wifi_station uses. This increases non-heap RAM from 48.5 kB to 80.5 kB.

TODO:
- [x] double check all samples


ncmiller (approved):
LGTM, assuming sample testing goes well

mniestroj:
All samples tested. There is an issue with esp32 platform and DFU sample (does not confirm and reboot automatically), but not related to this PR. Merging.

PR #286, opened 2022-09-16
Nick/mimxrt1060 log fix

ncmiller:


ncmiller:
This didn't work as intended, so closing.

PR #288, opened 2022-09-19
Release v0.3.0

mniestroj:
Notes:
- NCS was **not** bumped to the latest v2.1.0 yet. The reason is MBEDTLS_DTLS compatibility with TF-M and SPM changes in NCS.
- `samples/dfu/` is not reliable with `mimxrt1060_evkb` - it still results in crashes; problem seems to be some kind of race condition, because adding `LOG_INF()` statements in Zephyr init process allows to workaround that issue.


ncmiller (approved):
LGTM, nice work!

szczys (approved):
LGTM!

PR #290, opened 2022-09-23
samples: common: improvements in net_connect()

mniestroj:
Those improvements affect mostly `esp32` and `mimxrt1060_evkb` platforms, which require waiting for specific condition on network interface, before it can be used by `net_connect()` implementation.


samples/common/net_connect.c
```diff
@@ -12,10 +12,65 @@ LOG_MODULE_REGISTER(golioth_samples, LOG_LEVEL_DBG);
...
+	} else {
+		wait_for_iface_up_poll(iface);
+	}
+}
+
 void net_connect(void)
 {
 	struct net_if *iface = net_if_get_default();
 
+	wait_for_iface_up(iface);
```
ncmiller, line 89: Might be helpful to put a log message here, in case it gets stuck.

ncmiller (approved):
Thanks for the quick fixes!

samples/common/net_connect.c
mniestroj: Added `LOG_INF("Waiting for interface to be up");`. Is that fine?

ncmiller: Perfect, thanks

szczys (approved):
LGTM!

PR #295, opened 2022-09-29
bump NCS to v2.1.0

mniestroj:
NRF_SECURITY module in NCS is enabled by default and we probably want to
keep it enabled, so that existing NCS users are not restricted by Golioth
SDK (e.g. by forcing NRF_SECURITY to be disabled). This module seems to
dropped old/legacy SPM module and now uses something new, which pulls in
NCS fork of mbedTLS as a dependency. This is expressed by 'select
DISABLE_MBEDTLS_BUILTIN if MBEDTLS' in both NRF_SECURITY and
NRF_SECURITY_BACKEND Kconfig options.

Using NCS fork of mbedTLS means selecting
CONFIG_MBEDTLS_TLS_LIBRARY (CONFIG_MBEDTLS_TLS_LIBRARY is NCS fork,
CONFIG_MBEDTLS_BUILTIN is Zephyr fork of mbedTLS).

NCS fork (as compared to Zephyr fork) of mbedTLS has set of Kconfig options
that are named differently (like CONFIG_MBEDTLS_SSL_PROTO_DTLS in NCS vs
CONFIG_MBEDTLS_DTLS in Zephyr), while other options are named the same. As
a result we are selecting CONFIG_MBEDTLS_SSL_PROTO_DTLS and
CONFIG_MBEDTLS_GCM_C to match what net/golioth/Kconfig file is selecting
automatically for the user.

Disable unused CONFIG_MBEDTLS_KEY_EXCHANGE_DHE_PSK_ENABLED and
CONFIG_MBEDTLS_KEY_EXCHANGE_DHE_RSA_ENABLED. Those are enabled by default
in NCS, by questionable 'default y if !NET_L2_OPENTHREAD'. Leaving them in
default state (y) breaks the build because of unmet dependencies (other
mbedTLS configuration options).

Increase mbedTLS heap size from 10kB (in prj.conf) to 48kB in order to make
it run. 32kB was still not enough in case of samples/hello/, so 48kB was
chosen to give enough room for future.

Tested with `lightdb`, `lightdb_stream`, `logging` and `dfu` samples.

### DIFF
--- a/samples/dfu/boards/nrf9160dk_nrf9160_ns.conf
+++ b/samples/dfu/boards/nrf9160dk_nrf9160_ns.conf
@@ -21,5 +21,13 @@ CONFIG_NRF_MODEM_LIB=y
 # in poll() instead.
 CONFIG_GOLIOTH_SYSTEM_CLIENT_TIMEOUT_USING_POLL=y
 
-# Conflicts with MBEDTLS_DTLS
-CONFIG_BUILD_WITH_TFM=n
+# Use NCS fork of mbedTLS, as Zephyr built-in mbedTLS integration (CONFIG_MBEDTLS_BUILTIN) is in
+# conflict with CONFIG_NRF_SECURITY
+CONFIG_MBEDTLS_TLS_LIBRARY=y
+CONFIG_MBEDTLS_SSL_PROTO_DTLS=y
+CONFIG_MBEDTLS_GCM_C=y
+# ... and disable options y-selected by NCS for no good reason
+CONFIG_MBEDTLS_KEY_EXCHANGE_DHE_PSK_ENABLED=n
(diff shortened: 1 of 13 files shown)


ncmiller (approved):
> 10kB (in prj.conf) to 48kB 

That's a big jump, but I suppose there's not much we can do it about it right now.

mniestroj:
> > 10kB (in prj.conf) to 48kB
> 
> That's a big jump, but I suppose there's not much we can do it about it right now.

Not even able to easily measure peak consumption, as `mbedtls` shell command that I developed (and upstreamed to Zephyr) for that purpose works only with MBEDTLS_BUILTIN (Zephyr fork). Hence I only tried 32 (with failure) and then 48 (with success). So here it is. At least there is stil plenty of RAM (total of 200kB) to be utilized, so we are still far from the limit.

PR #297, opened 2022-10-05
settings: support for errors array in .c/status

ncmiller:
Errors for individual settings are now reported in the "errors" array of the settings status report, per the latest backend spec.

Increased the maximum response len from 64 to 256 since multiple errors can increase the message size, which would result in QCBOR encode failure.

### DIFF
--- a/net/golioth/settings.c
+++ b/net/golioth/settings.c
@@ -29,8 +29,11 @@
  * Example settings response from device:
  *
  * {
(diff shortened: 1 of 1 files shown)


ncmiller:
checkpatch fails because of QCBOR Camel Case (`Avoid CamelCase`), but this is a false positive and should not prevent this PR from merging.

PR #298, opened 2022-10-06
sockets_offload_poll_wrapper: introduce new module and use it by nRF91

mniestroj:
There is a special ZFD_IOCTL_POLL_OFFLOAD ioctl() request used inside
poll() system call handling, which allows to offload whole poll() system
call to driver offloading sockets operations (like NCS' nRF91 sockets
driver). Offloading whole poll() execution doesn't allow to use Zephyr
native sockets (either from another network driver or just software
concepts like socketpair and eventfd). This is a huge restriction, because
it prevents "interrupting" poll() system call and blocks thread calling
poll() until new data arrives (at least when using POLLIN event).

Introduce offloaded sockets wrapper, which wraps all socket operations with
special behavior added specifically in poll() system call handling.
ZFD_IOCTL_POLL_PREPARE ioctl() request is handled (instead of returning
-EXDEV by offloaded sockets drivers) and execution of offloaded poll() is
actually scheduled in helper (one per socket) thread. Potentially
infinite (in case of no data arriving on socket) execution of poll() just
makes the helper thread to sleep, while the application thread is not
affected. Application thread is still notified about POLLIN event, by
raising k_poll_signal by helper thread. Such k_poll_signal event
notification is added next to other native Zephyr notification
mechanisms (like semaphore, fifo, msgq, pipe) to k_poll() doing the
heavylifting in poll() implementation. This means that such wrapped poll()
handling allows to combine it with other Zephyr sockets, eliminating major
restriction of "offloaded poll" mechanism.

Limitations of current implementation:
 * only POLLIN event is handled (no POLLOUT event handling),
 * POLLERR and POLLHUP are not propagated.

Use of this module allows to utilize eventfd mechanism, which is used by
all other platforms. For now this change should not result in functional
changes (this is just implementation mechanism change). However 'eventfd'
will be the requirement with the upcoming CoAP rework.

### DIFF
--- a/net/sockets/sockets_offload_poll_wrapper.c
+++ b/net/sockets/sockets_offload_poll_wrapper.c
@@ -1,0 +1,474 @@
+/*
+ * Copyright (c) 2022 Golioth, Inc.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <zephyr/logging/log.h>
+#include <zephyr/net/socket.h>
+
+#include "sockets_internal.h"
+
+LOG_MODULE_REGISTER(net_sock_wrapper, CONFIG_NET_SOCKETS_LOG_LEVEL);
+
+__net_socket struct wrapper_context {
+	int fd;
+	int my_fd;
+	bool is_used;
+	struct k_poll_signal poll_signal;
+	struct k_work work;
+	struct k_work_q work_q;
+};
+
+static struct wrapper_context
+	wrapper_contexts[CONFIG_NET_SOCKETS_OFFLOAD_POLL_WRAPPER_CONTEXT_MAX];
+
+K_THREAD_STACK_ARRAY_DEFINE(sock_wrapper_poll_stack,
+			    CONFIG_NET_SOCKETS_OFFLOAD_POLL_WRAPPER_CONTEXT_MAX,
+			    CONFIG_NET_SOCKETS_OFFLOAD_POLL_WRAPPER_THREAD_STACK_SIZE);
+
+#define SOCK_WRAPPER_POLL_STACK_NAME(i, _) "threaded poll " #i
+
+static const char * const sock_wrapper_poll_stack_names[] = {
+	LISTIFY(CONFIG_NET_SOCKETS_OFFLOAD_POLL_WRAPPER_CONTEXT_MAX,
+		SOCK_WRAPPER_POLL_STACK_NAME, (,))
+};
+
+static K_MUTEX_DEFINE(wrapper_lock);
+
+static int sock_wrapper_create(int family, int type, int proto);
+
+static ssize_t sock_wrapper_recvfrom_vmeth(void *obj, void *buf,
+					   size_t max_len, int flags,
+					   struct sockaddr *addr,
+					   socklen_t *addrlen);
+
+static ssize_t sock_wrapper_sendto_vmeth(void *obj, const void *buf,
+					 size_t len, int flags,
+					 const struct sockaddr *addr,
+					 socklen_t addrlen);
+
+static void __wrapper_ctx_free(struct wrapper_context *wrapper)
+{
+	wrapper->fd = -1;
+	wrapper->my_fd = -1;
+	wrapper->is_used = false;
+}
+
+static int sock_wrapper_socket(int family, int type, int proto, bool *is_offloaded)
+{
+	STRUCT_SECTION_FOREACH(net_socket_register, sock_family) {
+		/* Ignore wrapper itself. */
+		if (sock_family->handler == sock_wrapper_create) {
+			continue;
+		}
+
+		if (sock_family->family != family &&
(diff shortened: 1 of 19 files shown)


net/sockets/sockets_offload_poll_wrapper.c
```diff
@@ -0,0 +1,474 @@
...
+
+	ARG_UNUSED(dev);
+
+	for (wrapper = wrapper_contexts;
+	     wrapper < &wrapper_contexts[ARRAY_SIZE(wrapper_contexts)];
+	     wrapper++) {
+		int idx = wrapper - wrapper_contexts;
+
+		k_poll_signal_init(&wrapper->poll_signal);
+		k_work_init(&wrapper->work, sock_wrapper_poll_handler);
```
ncmiller, line 457: Is it true that we will be creating one thread per `k_work_q`? With the default value in Kconfig, this would result in POSIX_MAX_FDS threads, or 16 threads (I think), which seems a bit wasteful.

mniestroj: It is true that there will be one thread per `k_work_q`. Those can be listed with `kernel threads` using shell if you want to try out and see yourself.

>this would result in POSIX_MAX_FDS threads, or 16 threads (I think), which seems a bit wasteful.

It is wasteful, yes. However default value of POSIX_MAX_FDS is 4, so there is slightly over 4kB (1024 bytes of stack plus overhead for stack overflow detection area) of stack space allocated when enabling this option.

mniestroj: Also see https://devzone.nordicsemi.com/f/nordic-q-a/76379/nrf9160-offloaded-poll-is-incompatible-with-other-zephyr-file-descriptors-e-g-eventfd for related DevZone request.

mniestroj: >>this would result in POSIX_MAX_FDS threads, or 16 threads (I think), which seems a bit wasteful.
>
>It is wasteful, yes. However default value of POSIX_MAX_FDS is 4, so there is slightly over 4kB (1024 bytes of stack plus overhead for stack overflow detection area) of stack space allocated when enabling this option.

This is a workaround, not a final solution. IMHO Nordic should fix their nRF91 sockets implementation, as it is both buggy and constrained.

ncmiller: Sounds good, thanks for the explanation.

ncmiller (approved):

lemrey:
Hello, haven't had a thorough look at this yet. It seems to address the fact that `poll` in Zephyr can't be used on both Zephyr sockets and nRF91 sockets. Do I understand you'd like to leverage `eventfd` and similar mechanisms to interrupt `poll()` calls on Zephyr and offloaded sockets alike? If so, would it help with your use-case if `nrf_poll` supported callbacks for the various `POLLIN` etc events, that could be set via a dedicated socket option? I guess that could be used the extend the code in the offloading implementation for nR91 to support your use case.

mniestroj:
> If so, would it help with your use-case if nrf_poll supported callbacks for the various POLLIN etc events, that could be set via a dedicated socket option? I guess that could be used the extend the code in the offloading implementation for nR91 to support your use case.

I am not sure what you mean by such callback. Could you provide some example (could be pseudo-code) that demonstrates that?

IMO here should be some mechanism (other than timeout that is passed as argument) to trigger `nrf_poll()` to finish it's execution before data is received/sent on nRF socket. Ideally, there should be some mechanism implemented in low-level nRF91 sockets drivers to implement Zephyr's `k_poll()`, so that nRF91 sockets would support Zephyr poll() and not the "offloaded" variant of poll() (which cannot wait on generic Zephyr sockets at the same time as it waits on nRF sockets events). Looking at the upstream Zephyr driver it looks like all the drivers (which use offloaded sockets) could actually implement Zephyr poll(), as they wait on a generic Zephyr object like `k_sem` or something similar. Right now the only exception is nRF91 sockets implementation, which *requires* offloaded poll() implementation and hence does not support passing other sockets (e.g. from other network interfaces, or eventfd file descriptors).

lemrey:
You could take a look at [this commit](https://github.com/lemrey/sdk-nrf/commit/1a4a525805097759d80d01ab7e13c61ab3b4ba3f) perhaps.

mniestroj:
It seems exactly what I need! Keep me posted when you have something and I'll test it with my use case.

Happy holidays!

lemrey:
Hello @mniestroj, it should be possible to do what you need now in `ncs/main`.

mniestroj:
@lemrey Thanks, I'll try it out soon!

mniestroj:
@lemrey Thanks for your development. Last week I tried NCS 2.3.0 and the way nRF91 sockets poll() is handled now is exactly what we needed. Works great, thanks once again!

PR #301, opened 2022-10-07
tests: lightdb: add basic LightDB GET tests

mniestroj:
Add tests that verify successful retrievement of LightDB values. For
simplicity just string values were used, so that there is no reordering
issue of keys within a structure (server can decide to return structures
with keys being ordered differently than what was set using goliothctl).

When application runs, it expects specific values at specific LightDB
paths. Those values are compared with values embedded as C arrays and
generated from resources/ directory. In order to make this test work
properly, those values can be automatically set using
scripts/golioth-setup.sh script, which uses goliothctl underneath.

The general idea is to run this test application from within Zephyr's
twister tool. In order to make it work, credentials and server
information (if self-hosted server is used) need to be provided to build
process. This is achieved by Kconfig project-specific default values, which
are fetched from environment variables if they exist.

At this point only qemu_x86 platform was tested when building and running
standalone, as well as from within twister. However, other platforms might
be supported as well, if network_setup() function (in main.c) would be
expanded to setup network.

This is a reopen of #273 (closed automatically by GitHub).


tests/lightdb/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+
+  add_custom_target(${generated_target_name} DEPENDS ${generated_file})
+
+  # Ensure 'generated_file' is generated before 'target' by creating a
+  # dependency between the two targets
+  add_dependencies(${target} ${generated_target_name})
+endfunction()
+
+foreach(name IN ITEMS small big-1024 big-2048)
+  set(path ${APPLICATION_SOURCE_DIR}/resources/get/${name}.json)
```
ncmiller, line 34: Unused variable `path`

ncmiller (approved):
LGTM. Interesting code generation stuff!

PR #304, opened 2022-10-17
coap_req: allow NULL callback

ncmiller:
Previously, the cb parameter of golioth_coap_req_cb was required. If it was set to NULL, the system would crash due to NULL dereference.

There have been a couple of instances of crashes which were due to this behavior. In one case, it was inadvertant (user meant to set a callback, but forgot). In the other case, it was intentional (rpc.c and settings.c both set the callback to NULL when sending the response, because they don't care about the response).

It seems like the most reasonable thing to do is to allow for a NULL callback, since there are legitimate requests that don't care to be called back ("fire and forget").

For cases where setting a NULL callback is unintentional, the user will be alerted soon enough when they realize that nothing is happening in response to their request (because they forget to set the callback). But at least there won't be a system crash (which can be hard to track down).

### DIFF
--- a/net/golioth/coap_req.c
+++ b/net/golioth/coap_req.c
@@ -354,6 +354,12 @@ void golioth_coap_req_process_rx(struct golioth_client *client, const struct coa
 	k_mutex_unlock(&client->coap_reqs_lock);
 }
 
+/* Default request callback, in case the caller does not specify one */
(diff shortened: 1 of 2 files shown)


mniestroj (approved):
Solves most important issues in a very non-intrusive way. Thanks a lot!

PR #315, opened 2022-10-27
samples: add 'rpc' testing and convert 'logging' to Golioth Python lib

mniestroj:
Add Golioth Python package that contains library with simple CLI tool to access Golioth services and
communicate with device (indirectly via cloud).

Add basic pytest runtime test for `rpc` samples, which tries to call RPC method `multiply` several
times and verify returned results.

Convert also `samples/logging/` pytest script to use introduced Python package, instead of calling
`goliothctl` subprocess. This shows clearly what are the differences of both solutions in test case
implementation.


scripts/python/golioth/golioth/golioth.py
```diff
@@ -0,0 +1,363 @@
...
+
+
+class ApiNodeMixin:
+    @property
+    def http_client(self):
+        return httpx.AsyncClient(base_url=self.base_url)
+
+    async def get(self, *args, **kwargs):
+        async with self.http_client as c:
+            return await c.get(*args, **kwargs)
```
ncmiller, line 66: It's okay to do in a future PR, but I would like to see support for using `https` with Golioth prod (or dev). It shouldn't be much more additional work, other than adding the `s` in the URL and setting the `x-api-key` header in request.

scripts/python/golioth/pyproject.toml
```diff
@@ -0,0 +1,35 @@
+[build-system]
```
ncmiller, line 1: Any pros/cons for keeping this package here vs in an external repo? I think it would be nice to put it in an external repo, but there might be reasons for keeping it here too.

This is a really great tool, so I want to make sure it's visible and available to external users.

samples/rpc/pytest/test_sample.py
```diff
@@ -0,0 +1,65 @@
...
+    start = datetime.now()
+
+    while datetime.now() < start + timedelta(seconds=initial_timeout):
+        try:
+            logging.debug('Calling "multiply" to check if device is connected')
+            await device.rpc.multiply(1, 1)
+
+            # No exception was raised, so just return device
+            return device
+        except RPCTimeout:
```
ncmiller, line 50: One thing I've been trying to capture lately (without success) is the latency of RPCs (min, average, max, etc). I suspect there's some kind of issue on the backend, since I sometimes see 2-3 second response times, but I haven't been able to prove it. I think this test (or something like it), could be used to call the RPC 1000 times and measure the results. I'm looking forward to it!

ncmiller (commented):
I'm really excited about this python package. Nice work!

scripts/python/golioth/README.md
```diff
@@ -0,0 +1,3 @@
+# Golioth client Python library
+
+This is a basic Golioth client library.
```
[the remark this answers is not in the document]
mniestroj: Okay, I'll update with package installation and basic CLI usage.

scripts/python/golioth/golioth/golioth.py
mniestroj: Sure. Didn't want to spend time on something "just in case", as I use 99% self-hosted server for testing. And this tool was developed for testing primarily, with other use-cases just kept in mind.

scripts/python/golioth/pyproject.toml
mniestroj: Initially this was supposed to be bunch of scripts in the repo, included by pytest scripts or alternatively called by used by providing full path (CLI). Including those scripts from pytest scripts would be hacky (due to `../../../` usage), so I decided to make an installable package out of it. This way there is the CLI tool available without providing full path and pytest scripts can just import the module just as other Python modules/packages.

The reason why NOT to put it in separate repo for now is due to maintenance costs. Keeping all in single repo allows to quickly adjust and break APIs. And as this only supports HTTP and self-hosted server, there are no concerns at all with breaking something. And the main use-case is testing as I said (either automatic or manual), so I didn't want to increase testing effort even more by putting the maintenance of "full fledged" package to our team.

samples/rpc/pytest/test_sample.py
mniestroj: Definitely. There is also `pytest-repeat` plug, which allows you to to something similar without even adjusting this test. There are other plugins for generating reports as well, which should also include such timing information.

scripts/python/golioth/pyproject.toml
ncmiller: Makes sense. We can keep it in the repo.

scripts/python/golioth/README.md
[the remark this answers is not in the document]
mniestroj: Updated!

scripts/python/golioth/README.md
```diff
@@ -0,0 +1,100 @@
...
+  --help              Show this message and exit.
+
+Commands:
+  call      Call RPC method on device.
+  logs      Logging service related commands.
+  settings  Settings service related commands.
+```
+
+or for each subcommand like `golioth logs --help`.
+
```
ncmiller, line 36: Can you mention the local configuration option you shared with me, about creating the `.golioth-local.yaml` file? Also, it's probably a good idea to add that file to `.gitignore`.

mniestroj:
```
golioth --help
```
mentions:
```
Options:
  --config-path PATH  Path to goliothctl configuration
```

ncmiller: Yeah, I was thinking of adding something just a convenience to people like me :)

For instance, I was getting a cryptic python error, and didn't know I should be using this option at all. I also was not familiar with what all is entailed in goliothctl configuration (e.g. the file type of YAML and the minimal config options I needed to set).

PR #328, opened 2022-12-08
coap_req: don't handle requests on network down

mniestroj:
First of all, do not allow to schedule CoAP requests anymore when client is
disconnected from Golioth server. From now on, whenever
`golioth_coap_req_cb()` API (and all APIs built on top of it, like LightDB
State APIs, LightDB Stream APIs, DFU, etc.) will be called in
"disconnected" state, scheduling such CoAP request will fail with `-ENETDOWN`
error code. This will let higher layers know that communication with server
is not possible and either such request can be dropped immediately (for
synchronous communication, like responding to RPC) or enqueued for sending
when connection is brought back.

Cancel all pending requests on disconnect event, so that those requests do
not wait infinitely for new connection with Golioth server. Instead, invoke
user-defined callbacks with `-ESHUTDOWN` error code, so that such "request
cancelled" event can be handled accordingly.

Add API tests when there is "connected" and "disconnected" state of Golioth
client. Check expected error codes and payloads (when successful) in both
states.

Check also behavior of callback-based API when the request was sent during
"connected" state and client got disconnected immediately after. Such case
should result in `-ESHUTDOWN` due to request being cancelled on
"disconnected" event.


net/golioth/coap_req.c
```diff
@@ -731,3 +750,47 @@ int64_t golioth_coap_reqs_poll_prepare(struct golioth_client *client, int64_t no
...
+{
+	struct golioth_coap_req *req, *next;
+
+	SYS_DLIST_FOR_EACH_CONTAINER_SAFE(&client->coap_reqs, req, next, node) {
+		struct golioth_req_rsp rsp = {
+			.user_data = req->user_data,
+			.err = reason,
+		};
+
+		(void)req->cb(&rsp);
```
ncmiller, line 765: Should we check that the callback is not `NULL` first?

ncmiller (commented):
Looks good, just one question

net/golioth/coap_req.c
mniestroj: No, we don't. `req->cb` is always initialized to some value. If user passes `NULL`, then `golioth_req_rsp_default_handler` is set in `golioth_coap_req_init()`.

ncmiller (approved):

PR #330, opened 2022-12-09
Kconfig: add GOLIOTH_SETTINGS_MAX_RESPONSE_LEN

ncmiller:
Before this change, the max length of the settings response was fixed at 256 (#define'd in the source code of settings.c). This was problematic for devices with many settings, because if there are multiple errors encountered while handling the settings, the response length becomes too large to fit in the response buffer, resulting in a QCBOR error:

QCBOREncode_FinishGetSize error: 1 (QCBOR_ERR_BUFFER_TOO_SMALL)

To accomodate users with lots of settings, make the response length configurable via Kconfig.

The recommendation of 50 * N for the max length is given because each error encountered results in a CBOR entry like:

{ "setting_key": string, "error_code": int }

The "setting_key" and "error_code" strings alone account for 21 bytes, then if you add in the name of the setting (string) and 1-byte error code, it can easily be 40-50 bytes for each error.

### DIFF
--- a/net/golioth/settings.c
+++ b/net/golioth/settings.c
@@ -47,12 +47,11 @@ LOG_MODULE_DECLARE(golioth);
 #define GOLIOTH_SETTINGS_PATH ".c"
 #define GOLIOTH_SETTINGS_STATUS_PATH ".c/status"
 #define GOLIOTH_SETTINGS_MAX_NAME_LEN 63 /* not including NULL */
-#define GOLIOTH_SETTINGS_MAX_RESPONSE_LEN 256
 
 struct settings_response {
 	QCBOREncodeContext encode_ctx;
 	UsefulBuf useful_buf;
-	uint8_t buf[GOLIOTH_SETTINGS_MAX_RESPONSE_LEN];
+	uint8_t buf[CONFIG_GOLIOTH_SETTINGS_MAX_RESPONSE_LEN];
 	size_t num_errors;
 };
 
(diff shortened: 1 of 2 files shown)


net/golioth/settings.c
```diff
@@ -112,6 +111,10 @@ static int finalize_and_send_response(struct golioth_client *client,
 	QCBORError qerr = QCBOREncode_FinishGetSize(&response->encode_ctx, &response_len);
 
 	if (qerr != QCBOR_SUCCESS) {
+		/*
+		 * Note: If the error is 1 (QCBOR_ERR_BUFFER_TOO_SMALL), then try increasing
```
mniestroj, line 115: We don't have any influence on the numbering of error codes in QCBOR, so I'd rather drop the reference for `1` and just leave `QCBOR_ERR_BUFFER_TOO_SMALL`.

One thing to consider here is to log a specific error message on `QCBOR_ERR_BUFFER_TOO_SMALL` error (with another `if` or with `switch`), so that instead of generic `QCBOREncode_FinishGetSize error: 1 (QCBOR_ERR_BUFFER_TOO_SMALL)` there will be something like `Settings response too large (consider increasing GOLIOTH_SETTINGS_MAX_RESPONSE_LEN)`.

net/golioth/settings.c
```diff
@@ -111,6 +110,13 @@ static int finalize_and_send_response(struct golioth_client *client,
 	size_t response_len;
 	QCBORError qerr = QCBOREncode_FinishGetSize(&response->encode_ctx, &response_len);
 
+	if (qerr == QCBOR_ERR_BUFFER_TOO_SMALL) {
+		LOG_ERR(
+			"Settings response too large "
```
mniestroj, line 115:
```suggestion
		LOG_ERR("Settings response too large "
```

PR #335, opened 2023-01-13
net: add log message for invalid format

szczys:
Poorly formatted JSON is a common cause for errors when using LightDB State/Stream. This commit watches for the -EFAULT error code and publishes a warning to check the formatting.

### DIFF
--- a/net/golioth/coap_req.c
+++ b/net/golioth/coap_req.c
@@ -204,6 +204,10 @@ static int golioth_coap_req_reply_handler(struct golioth_coap_req *req,
 	LOG_DBG("CoAP response code: 0x%x (class %u detail %u)",
 		(unsigned int)code, (unsigned int)(code >> 5), (unsigned int)(code & 0x1f));
 
+	if (code == COAP_RESPONSE_CODE_BAD_REQUEST) {
+		LOG_WRN("Server reports CoAP Bad Request. (Check payload formatting)");
+	}
+
 	err = golioth_coap_code_to_posix(code);
 	if (err) {
 		struct golioth_req_rsp rsp = {


net/golioth/coap_req.c
```diff
@@ -629,7 +629,12 @@ int golioth_coap_req_sync(struct golioth_client *client,
 	k_sem_take(&sync_data.sem, K_FOREVER);
 
 	if (sync_data.err) {
-		LOG_WRN("req_sync finished with error %d", sync_data.err);
+		if (sync_data.err == -EFAULT) {
+			LOG_WRN("Server reports CoAP Bad Request. (Check payload formatting)");
```
ncmiller, line 633: Since this is an error reported by the server, it might make more sense to put this warning in the coap reply callback, before it gets converted to EFAULT:

https://github.com/golioth/golioth-zephyr-sdk/blob/main/net/golioth/coap_req.c#L206

Something like:

```c
if (code == COAP_RESPONSE_CODE_BAD_REQUEST) {
    LOG_WRN("Server reports CoAP Bad Request. (Check payload formatting)");
}
```

szczys: That's a much better place for it. I've made the change, thanks!

net/golioth/coap_req.c
```diff
@@ -204,6 +204,10 @@ static int golioth_coap_req_reply_handler(struct golioth_coap_req *req,
 	LOG_DBG("CoAP response code: 0x%x (class %u detail %u)",
 		(unsigned int)code, (unsigned int)(code >> 5), (unsigned int)(code & 0x1f));
 
+	if (code == COAP_RESPONSE_CODE_BAD_REQUEST) {
+	    LOG_WRN("Server reports CoAP Bad Request. (Check payload formatting)");
```
ncmiller, line 208: Mix of tab and spaces is causing checkpatch to fail.

szczys: Apologies, checks have now passed.

I've added checkpatch to my pre-commit hooks to (hopefully) avoid this issue in the future.

ncmiller (approved):
Looks good, thanks!

PR #337, opened 2023-02-03
rpc, settings: remove automatic observations

ncmiller:
Currently, when golioth_rpc_register() is called, an observation is automatically established (if it’s the first RPC method being registered).

Similarly, when golioth_settings_register_callback() is called, an observation is automatically established.

However, this “automatic observe” behavior has three main issues:

1. It might be unwanted in some user applications (e.g. low-power/low-data) that don’t want the overhead of observations

2. It’s inconsistent with the way observations work for DFU. For DFU, it’s expected that the user calls golioth_fw_observe_desired() in the on_connect callback (i.e. observations are not automatically established).

3. It can be unexpected/surprising that an API named “register callback” would also create an observation.

We should separate RPC and Settings registration from observation, and make the observation an explicit API call.

This is a breaking change, since user applications will need to be modified to call the new functions in the on_connect callback.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,18 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.6.0] - TBD
+### Breaking Changes
+- calling `golioth_rpc_register()` no longer automatically observes. The RPC observation is now
+  established only if the user calls `golioth_rpc_observe()`. Existing applications that use
+  RPCs will need to call `golioth_rpc_observe()` in the `on_connect` callback. Additionally, it's
+  recommended to move any calls to `golioth_rpc_register()` outside of the `on_connect` callback,
+  to avoid registering the same RPC method multiple times. See the `rpc` sample for reference.
+- calling `golioth_settings_register_callback()` no longer automatically observes. The settings
+  observation is now established only if the user calls `golioth_settings_observe()`.
+  Existing applications that use settings will need to call `golioth_settings_observe()` in the
+  `on_connect` callback. See the `settings` sample for reference.
+
 ## [0.5.0] - 2022-12-13
 ### Added
 - Python library and CLI tools for accessing Golioth REST API (`scripts/python/golioth/`)
--- a/include/net/golioth/rpc.h
+++ b/include/net/golioth/rpc.h
@@ -140,6 +140,24 @@ int golioth_rpc_register(struct golioth_client *client,
 			 golioth_rpc_cb_fn callback,
 			 void *callback_arg);
 
+/**
+ * @brief Observe for RPC method invocations
+ *
+ * User applications should call this function in the `on_connect` callback,
+ * if they wish to observe RPCs.
+ *
+ * Establishes a single observation for endpoint ".rpc".
+ * The handler for this endpoint will look up the method in a table of
+ * registered RPCs (from @ref golioth_rpc_register) and invoke the callback
+ * if the method is found.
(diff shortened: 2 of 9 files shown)


include/net/golioth/rpc.h
```diff
@@ -129,6 +129,24 @@ int golioth_rpc_register(struct golioth_client *client,
 			 golioth_rpc_cb_fn callback,
 			 void *callback_arg);
 
+/**
+ * @brief CoAP Observe for RPC method invocations
```
mniestroj, line 133: There is nothing CoAP specific about it from user perspective.

include/net/golioth/settings.h
```diff
@@ -128,6 +128,23 @@ struct golioth_settings {
 int golioth_settings_register_callback(struct golioth_client *client,
 				       golioth_settings_cb callback);
 
+/**
+ * @brief CoAP Observe for Settings
```
mniestroj, line 132: Same here, there is nothing CoAP specific about this API.

samples/rpc/src/main.c
```diff
@@ -60,6 +57,13 @@ void main(void)
 	client->on_connect = golioth_on_connect;
 	golioth_system_client_start();
 
+	assert(IS_ENABLED(CONFIG_GOLIOTH_RPC));
```
mniestroj, line 60: I'm not fan of this `assert()`, as it bloats code just to double check if the `prj.conf` configuration option is applied. If it won't be set, then there will be a warning in cmake configuration step already.

net/golioth/rpc.c
```diff
@@ -189,9 +189,6 @@ int golioth_rpc_register(struct golioth_client *client,
 	method->callback_arg = callback_arg;
 
 	client->rpc.num_methods++;
-	if (client->rpc.num_methods == 1) {
-		status = golioth_rpc_observe(client);
-	}
```
mniestroj, line 194: If user will call `golioth_rpc_observe()` alone or after `golioth_rpc_register()`, then there is a chance of using `client->rpc_mutex` without first initializing it (as it happens in `golioth_rpc_register()`). Let's move this mutex initialization to be part of `golioth_init()` (but with some helper function like `golioth_rpc_init()`). Also, it would be good idea to move `struct k_mutex rpc_mutex;` into `struct golioth_rpc rpc;`.

CHANGELOG.md
```diff
@@ -4,6 +4,18 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.6.0] - TBD
+### Breaking Changes
+- Calling `golioth_rpc_register()` no longer automatically observes. The RPC observation is now
```
mniestroj, line 9:
```suggestion
- calling `golioth_rpc_register()` no longer automatically observes. The RPC observation is now
```
start with lowercase, just to be consistent with other released

CHANGELOG.md
```diff
@@ -4,6 +4,18 @@ All notable changes to this project will be documented in this file.
...
+  established only if the user calls `golioth_rpc_observe()`. Existing applications that use
+  RPCs will need to call `golioth_rpc_observe()` in the `on_connect` callback. Additionally, it's
+  recommended to move any calls to `golioth_rpc_register()` outside of the `on_connect` callback,
+  to avoid registering the same RPC method multiple times. See the `rpc` sample for reference.
+- Calling `golioth_settings_register_callback()` no longer automatically observes. The settings
```
mniestroj, line 14:
```suggestion
- calling `golioth_settings_register_callback()` no longer automatically observes. The settings
```
start with lowercase, just to be consistent with other released

mniestroj (approved):

PR #340, opened 2023-02-14
samples: common: introduce lte_monitor for nRF91

mniestroj:
It is hard to know what is the reason of nRF91 taking sometimes so long to
boot. As the LTE registration/attachment happens all in Zephyr OS boot
phase (before `main()` is executed), there are no logs printed and nothing
really indicates what is the reason of hang.

Add LTE monitor module specifically for nRF91 modem. This will log events
related to LTE Link Control, such as:

```
<inf> lte_monitor: Network: Searching
```

and after connecting:

```
<inf> lte_monitor: RRC: Connected
<inf> lte_monitor: Network: Registered (roaming)
```

which will give user a hint about what happens and why doesn't the
application continue to run as expected.

This also should catch all the disconnections when device has weak LTE
signal and give a hint about LTE being the reason of disconnection (or
reconnection) from Golioth.


samples/common/nrf91_lte_monitor.c
```diff
@@ -0,0 +1,65 @@
...
+		case LTE_LC_NW_REG_REGISTERED_ROAMING:
+			LOG_INF("Network: Registered (roaming)");
+			break;
+		case LTE_LC_NW_REG_REGISTERED_EMERGENCY:
+			LOG_INF("Network: Registered (emergency)");
+			break;
+		case LTE_LC_NW_REG_UICC_FAIL:
+			LOG_INF("Network: UICC fail");
+			break;
+		}
```
ncmiller, line 42: Missing a `break` here.

ncmiller (commented):
Nice quality-of-life improvement

samples/common/nrf91_lte_monitor.c
mniestroj: Good catch!

ncmiller (approved):

PR #345, opened 2023-02-21
bump NCS to 2.3.0

mniestroj:
Update to the latest release candidate to keep track of the upstream
improvements and changes.


mniestroj:
So far there is no strong reason to merge, so let's wait for 2.3.0 before making any progress on that one.

PR #348, opened 2023-03-07
golioth: fix zsock_setsockopt(..., TLS_HOSTNAME, ...) API use

mniestroj:
`zsock_setsockopt()` expects to receive `NULL` terminated string and length of
the string **including** `NULL` termination. Adjust code to follow that.

See Zephyr upstream commit 4234e801b544 ("net: sockets: Fix TLS_HOSTNAME
option length inconsistency") for details.


mniestroj:
```
aaa91be5860fc66dc10aad6d36b9daa6b44e1325:9: WARNING:UNKNOWN_COMMIT_ID: Unknown commit id '4234e801b544', maybe rebased or not pulled?
```
is a false-positive in this context. This commit sha1 is from Zephyr upstream, not from this repo. It is explicitly stated so.

PR #350, opened 2023-03-16
scripts: python: golioth: handle generic logs with Zephyr formatting

mniestroj:
There is no `uptime` field when generic logs are sent by cloud, such as
notifications about DFU state. Handle that separately by reusing default
log format and prepending those with `GENERIC: ` string.


szczys (approved):
LGTM!

PR #351, opened 2023-03-16
west-zephyr: import all Zephyr modules

mniestroj:
In case of `west-zephyr.yml` there is a name-allowlist filter, while in case
of `west-ncs.yml` there is no such filter. This makes filtering logic
inconsistent across Zephyr vanilla vs NCS.

Limiting repositories has a major flaw when user wants to utilize a generic
library such as picolibc, which is filtered out. In that case it is needed
to add that repository to downstream manifest or temporarily modify
`west-zephyr.yml` in order to test some non-standard configuration with SDK
samples.

Let's be consistent about Zephyr modules filtering and always import all
the modules in `west-zephyr.yml`, just like it happens with `west-ncs.yml`.

Since limiting `west-zephyr.yml` was mainly a clone time improvement,
noticeable in CI, add .ci-`west-zephyr.yml` which will limit repositories
just for being run in CI.

### DIFF
--- a/.ci-west-zephyr.yml
+++ b/.ci-west-zephyr.yml
@@ -1,0 +1,17 @@
+manifest:
+  self:
+    path: modules/lib/golioth
+    import:
+      file: west-zephyr.yml
+      name-allowlist:
+        - cmsis
+        - hal_espressif
+        - hal_nordic
+        - hal_nxp
+        - mbedtls
+        - mcuboot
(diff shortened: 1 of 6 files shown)


szczys (approved):
LGTM!

I've run into this issue of standard modules that didn't get installed because of the allow list and this is a nice improvement to avoid that with newer users.

PR #352, opened 2023-03-16
release: v0.6.0

mniestroj:
See CHANGELOG.md for details.

Depends on:
- https://github.com/golioth/golioth-zephyr-sdk/pull/350
- https://github.com/golioth/golioth-zephyr-sdk/pull/351
- https://github.com/golioth/golioth-zephyr-sdk/pull/353

PR #353, opened 2023-03-17
samples: dfu: override esp_clk_init() with empty definition

mniestroj:
There is an issue when booting new firmware for the first time. This
results in boot hang. It seems that the reason is `esp_clk_init()`.
Fortunately this function definition is flagged as weak, so override it in
sample to make DFU functional.


mniestroj:
Already merged as part of #352

PR #354, opened 2023-03-22
Add asyncronous firmware report API

szczys:
* fw: add golioth_fw_report_state_cb() API
* samples: dfu: add golioth_fw_report_state_db()

This PR adds an asynchronous API that is safe to call from the system_client thread. This way the firmware state report can be made from the `golioth_on_connect()` callback.

### DIFF
--- a/net/golioth/fw.c
+++ b/net/golioth/fw.c
@@ -179,16 +179,14 @@ free_req:
 	return err;
 }
 
-int golioth_fw_report_state(struct golioth_client *client,
-			    const char *package_name,
-			    const char *current_version,
-			    const char *target_version,
-			    enum golioth_fw_state state,
-			    enum golioth_dfu_result result)
+static int fw_report_state_encode(const char *current_version,
+				  const char *target_version,
+				  enum golioth_fw_state state,
+				  enum golioth_dfu_result result,
+				  UsefulBuf encode_bufc,
+				  size_t *encoded_len)
 {
 	QCBOREncodeContext encode_ctx;
-	UsefulBuf_MAKE_STACK_UB(encode_bufc, 64);
-	size_t encoded_len;
 	QCBORError qerr;
 
 	QCBOREncode_Init(&encode_ctx, encode_bufc);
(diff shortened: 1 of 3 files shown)


samples/dfu/src/main.c
```diff
@@ -146,9 +146,40 @@ static int golioth_desired_update(struct golioth_req_rsp *rsp)
...
+	}
+
+	LOG_INF("Async firmware report succeeded");
+	return 0;
+}
+
 static void golioth_on_connect(struct golioth_client *client)
 {
 	int err;
+	static bool first_contact = true;
```
mniestroj, line 163: Please, no `static` variables within function body.

net/golioth/fw.c
```diff
@@ -213,10 +216,53 @@ int golioth_fw_report_state(struct golioth_client *client,
...
+			       enum golioth_dfu_result result,
+			       golioth_req_cb_t cb, void *user_data)
+{
+	return fw_report_state(client,
+			       package_name,
+			       current_version,
+			       target_version,
+			       state,
+			       result,
+			       (cb ? cb : golioth_req_rsp_default_handler),
```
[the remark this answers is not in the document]
szczys: I implemented the first bullet point you suggested.

I agree with the second bullet point, but it felt like a different effort than this current task. I would like to discuss looking at that next if you think it's a useful improvement.

samples/dfu/src/main.c
szczys: Thanks for catching this one. I reimplemented this using a semaphore. Please take a look and let me know if this is an okay approach.

mniestroj:
@szczys I've merged first commit, which introduces async API. Second commit is still there. The issue I have with it is that it changes sync to async API use, so one is tested instead of the other. I think for this sample sync API makes more sense.

szczys:
> The issue I have with it is that it changes sync to async API use, so one is tested instead of the other.

This was also a concern of mine that both the sync and async be tested. In this commit, I have changed the initial firmware report to async, but sync is still used when reporting the other other states (`DOWNLOADING`, `DOWNLOADED`, `UPDATING`).

The issue I have with this change is that it makes what is already a somewhat complex sample and adds more complexity, really just for the purpose of testing the async function. (It also demonstrates the function, but my intent was to include the new code into the existing testing routine.)

Maybe we should think about automated OTA testing (Hardware in the Loop via CI/CD) and move this change there?

mniestroj:
> The issue I have with this change is that it makes what is already a somewhat complex sample and adds more complexity,

Agree. Let's leave this sample minimal (and even simplify if possible).

> Maybe we should think about automated OTA testing (Hardware in the Loop via CI/CD) and move this change there?

Definitely a good idea. We have a LightDB specific tests application, which invokes various APIs for that service: https://github.com/golioth/golioth-zephyr-sdk/tree/main/tests/lightdb. An equivalent for DFU would be great. In such dedicated application we can focus on using and testing all APIs, without keeping it easy for first-time users to understand.

PR #356, opened 2023-04-03
samples: rpc: log during method invocations

mniestroj:
Right now this sample was silent on the Zephyr console, so the only way to
tell whether RPC was executed was using REST API and/or web console. This
however does not tell whether tested board was the one responding with
correct results.

Add debug logs, so that every RPC invocation results in clear notification
on Zephyr console.


szczys (approved):
LGTM!

PR #357, opened 2023-04-03
samples: hello: increase NET_BUF_TX_COUNT in Kconfig

mniestroj:
`NET_BUF_TX_COUNT` was increased in `nrf52840dk_nrf52840` overlay configuration
file. Do this in Kconfig by defining default value. This has following
advantages:
 * only applied with cert-based authentication (where original issue is
   reproducible)
 * applied for all esp-at boards (there is only one supported by board
   overlay, but potentially all ESP-AT boards might benefit from this
   change due to low-bandwidth nature of this WiFi interface)
 * all board overlay files across all samples are the same (at least for
   `nrf52840dk_nrf52840` board)


szczys (commented):
LGTM

PR #362, opened 2023-04-13
samples: lightdb_stream: remove explicit CONFIG_EVENTFD=y

mniestroj:
`CONFIG_EVENTFD` mechanism is selected automatically by `CONFIG_GOLIOTH`, so no
reason to do it explicitly in this sample.


szczys (approved):
LGTM! I tested and `CONFIG_EVENTFD=y` is selected without including it in the prj.conf.

PR #363, opened 2023-04-13
coap_req: fix log format for 'size_t'

mniestroj:
Use `%zu` instead of `%d` for printing `size_t new_offset` variable. This
suppresses build time warnings for `qemu_x86_64` platform:

```
coap_req.c:257:33: warning: format '%d' expects argument of type 'int', \
  but argument 2 has type 'size_t' {aka 'long unsigned int'} [-Wformat=]

```


szczys (approved):
LGTM! I tested out the `%zu` formatting and had no issues.

PR #365, opened 2023-04-20
golioth: remove 'client->rx_options'

mniestroj:
This member was never used. Since `coap_packet_parse()` handles passing `NULL`
pointer, use that instead of storing received CoAP options.


szczys (approved):
LGTM

PR #368, opened 2023-04-20
samples: convert pytest script to Python library

mniestroj:
All the Python library interfaces are complete in terms of requirements for pytest testing. This
makes it a perfect time to covert all existing pytest test cases to actually use that library
instead of executing `goliothctl` in a subprocess. This gives much more flexibility on filtering
data and timeout as well al error handling.


szczys (approved):
LGTM!

PR #369, opened 2023-04-20
Add DFU testing workflow

szczys:
This PR adds a workflow to test `samples/dfu` using Twister on the nRF52840dk via a self-hosted runner.

### DIFF
--- a/.github/workflows/twister_dfu_nrf52840dk.yml
+++ b/.github/workflows/twister_dfu_nrf52840dk.yml
@@ -124,6 +124,8 @@ jobs:
           source env_dfu_nrf52840dk.sh
           PATH=$PATH:/opt/nrf-command-line-tools/bin
           nrfjprog -i
+          golioth releases delete '*'
+          golioth artifacts delete '*'
           zephyr/scripts/twister -vvv -p nrf52840dk_nrf52840 -T modules/lib/golioth/samples/dfu --device-testing --hardware-map hardware-map.yaml --west-flash=''
 
       - name: Upload artifacts
--- a/.github/workflows/twister_dfu_nrf52840dk.yml
+++ b/.github/workflows/twister_dfu_nrf52840dk.yml
@@ -1,0 +1,136 @@
+name: Twister DFU nrf52840dk
+
+on:
+  push:
+    branches: [ main ]
+  schedule:
+    # Run workflow at the start of every day (12 AM UTC)
+    - cron: "0 0 * * *"
+
+  workflow_dispatch:
+
+jobs:
+  # Assumptions made about the self-hosted runner:
+  #
+  #  1. Has installed the GitHub Actions self-hosted runner service
+  #  2. Has nrfjprog already installed at /opt/nrf-command-line-tools
+  #  3. Has SEGGER tools already installed at /home/golioth/JLink_Linux_V786_arm64
+  #  4. GitHub secret YOUR_RUNNER_NAME_NRF52840DK_PORT
+  #  5. GitHub secret YOUR_RUNNER_NAME_NRF52840DK
+  #  6. GitHub secret GOLIOTHCTL_YAML
+  #
+  # It is the responsibility of the self-hosted runner admin to ensure
+  # these pre-conditions are met.
+  #
+  # For item 1, GitHub provides instructions when you add a new self-hosted runner
+  # in Settings -> Actions -> Runners.
+  #
+  # For item 2, you'll need to install jlink and nrf-command-line-tools:
+  #   https://www.segger.com/downloads/jlink/
+  #   https://www.nordicsemi.com/Products/Development-tools/nrf-command-line-tools/download
+  #
+  # For item 3, this environment variable can be added to $HOME/runner_env.sh:
+  #   export CI_NRF52840DK_PORT=/dev/ttyACM0
+  #
+  # For item 4, the file needs to have contents like (use base64 encoding):
+  #
+  # - connected: true
+  #   id: "001050257254"
+  #   platform: nrf52840dk_nrf52840
+  #   product: J-Link
+  #   runner: nrfjprog
+  #   serial: /dev/serial/by-id/usb-SEGGER_J-Link_001050257254-if00
+  #
+  # For item 5, the file needs to have contents like (use base64 encoding):
+  #
+  # export GOLIOTH_SYSTEM_SERVER_HOST=coap.golioth.dev
+  # export GOLIOTH_SYSTEM_CLIENT_PSK_ID=device_psk_id@ci
+  # export GOLIOTH_SYSTEM_CLIENT_PSK=device_psk
+  # export GOLIOTH_SAMPLE_WIFI_SSID=golioth-runner-xxx
+  # export GOLIOTH_SAMPLE_WIFI_PSK=password_for_wifi
+  # export GOLIOTH_DEVICE_NAME=nrf52840dk
+  #
+  # For item 6, the file needs to have contents like (use base64 encoding):
+  #
+  # accesstoken: veryverylongtokenhash
+  # apikey: ""
+  # apiurl: https://api.golioth.dev
+  # config: ""
+  # projectid: ci
+  #
+  # NOTE: To store files as GitHub secrets, use base64 encoding:
+  #
+  # $ base64 -i --wrap=0 your_secret_file.yaml
+
+  twister_DFU:
+    runs-on: [self-hosted, has_twister_nrf52840dk]
+
+    container:
+      image: zephyrprojectrtos/ci:v0.26.1
+      volumes:
+        - /dev:/dev
+        - /opt/nrf-command-line-tools:/opt/nrf-command-line-tools
+        - /opt/SEGGER/JLink:/opt/SEGGER/JLink
+      options: --privileged
+
+    env:
+      ZEPHYR_SDK_INSTALL_DIR: /opt/toolchains/zephyr-sdk-0.16.0
+
+    defaults:
+      run:
+        shell: bash
+        working-directory: twister_run
+
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          path: twister_run/modules/lib/golioth
+
+      - name: Init and update west
+        run: |
+          mkdir -p .west
+          cat <<EOF > .west/config
+          [manifest]
+          path = modules/lib/golioth
+          file = .ci-west-zephyr.yml
+          EOF
+
+          west update -o=--depth=1 -n
+          west patch --apply
+
+      - name: Installing Python reqs
+        run: |
+          cd modules/lib/golioth/
+          pip install -r scripts/requirements-tests.txt
+          cd scripts/python/golioth
+          pip install .
+
+      - name: Format Runner Name
+        run: |
+          RUNNERNAME=$(echo ${{ runner.name }} | sed -e 's/-/_/g' -e 's/[a-z]/\U&/g')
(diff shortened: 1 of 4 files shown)


.github/workflows/test_dfu_nrf52840dk.yml
```diff
@@ -0,0 +1,125 @@
...
+
+          west update -o=--depth=1 -n
+          west patch --apply
+
+      - name: Build samples/dfu
+        run: |
+          cd modules/lib/golioth/samples/dfu
+          west build -b nrf52840dk_nrf52840 . -p
+          west sign -t imgtool --no-hex -B build/golioth-v0.0.0.bin -- --key ../../../../../bootloader/mcuboot/root-rsa-2048.pem
+          west sign -t imgtool --no-hex -B build/golioth-v2.0.0.bin -- --key ../../../../../bootloader/mcuboot/root-rsa-2048.pem --version 2.0.0
```
mniestroj, line 45:
```suggestion
          west sign -t imgtool --no-hex -B build/golioth-v0.0.0.bin -- --key $(west topdir)/bootloader/mcuboot/root-rsa-2048.pem
          west sign -t imgtool --no-hex -B build/golioth-v2.0.0.bin -- --key $(west topdir)/bootloader/mcuboot/root-rsa-2048.pem --version 2.0.0
```

.github/workflows/twister_dfu_nrf52840dk.yml
```diff
@@ -0,0 +1,138 @@
...
+  #   export CI_NRF52840DK_PORT=/dev/ttyACM0
+  #
+  # For item 4, the file needs to have contents like (use base64 encoding):
+  #
+  # - connected: true
+  #   id: "001050257254"
+  #   platform: nrf52840dk_nrf52840
+  #   product: J-Link
+  #   runner: nrfjprog
+  #   serial: /dev/serial/by-id/usb-SEGGER_J-Link_001050257254-if00
```
mniestroj, line 42: This is runner specific, so IMO it will be better to store it on the machine connected to DUT, e.g. at `/etc/hardware-map.yaml` and then mapped into the container with another volume:
```
      volumes:
        - /etc/hardware-map.yaml:/etc/hardware-map.yaml
```

.github/workflows/twister_dfu_nrf52840dk.yml
```diff
@@ -0,0 +1,138 @@
...
+  #
+  # For item 5, the file needs to have contents like (use base64 encoding):
+  #
+  # export GOLIOTH_SYSTEM_SERVER_HOST=coap.golioth.dev
+  # export GOLIOTH_SYSTEM_CLIENT_PSK_ID=device_psk_id@ci
+  # export GOLIOTH_SYSTEM_CLIENT_PSK=device_psk
+  # export GOLIOTH_SAMPLE_WIFI_SSID=golioth-runner-xxx
+  # export GOLIOTH_SAMPLE_WIFI_PSK=password_for_wifi
+  # export GOLIOTH_DEVICE_NAME=nrf52840dk
+  # export GOLIOTHCTL_CONFIG=$HOME/.goliothctl-local.yaml
```
mniestroj, line 52: This is not needed if the file will be located at `$HOME/.golioth/.goliothctl.yaml` (this is the default path).

.github/workflows/twister_dfu_nrf52840dk.yml
```diff
@@ -0,0 +1,138 @@
...
+          echo ${{ secrets.GOLIOTHCTL_YAML }} | base64 -d > .goliothctl-local.yaml
+
+      - name: Run Twister
+        run: |
+          source env_dfu_nrf52840dk.sh
+          cp .goliothctl-local.yaml ${HOME}/.
+          sed -i 's/CONFIG_SETTINGS=y/CONFIG_SETTINGS=n/' modules/lib/golioth/samples/dfu/prj.conf
+          PATH=$PATH:/opt/nrf-command-line-tools/bin
+          nrfjprog -i
+          zephyr/scripts/twister -vvv -p nrf52840dk_nrf52840 -T modules/lib/golioth/samples/dfu --device-testing --hardware-map hardware-map.yaml --west-flash=''
```
mniestroj, line 129: We can provide `modules/lib/golioth/samples` to test all samples, not just `samples/dfu`. In that case it would be good to provide `-t goliothd`, as `goliothd` tag is used in all samples that require communication with Golioth backend.

.github/workflows/twister_dfu_nrf52840dk.yml
```diff
@@ -0,0 +1,138 @@
...
+        run: |
+          echo ${{ secrets[env.hil_runner_nrf52840dk] }} | base64 -d > env_dfu_nrf52840dk.sh
+          echo ${{ secrets[env.hil_runner_nrf52840dk_port] }} | base64 -d > hardware-map.yaml
+          echo ${{ secrets.GOLIOTHCTL_YAML }} | base64 -d > .goliothctl-local.yaml
+
+      - name: Run Twister
+        run: |
+          source env_dfu_nrf52840dk.sh
+          cp .goliothctl-local.yaml ${HOME}/.
+          sed -i 's/CONFIG_SETTINGS=y/CONFIG_SETTINGS=n/' modules/lib/golioth/samples/dfu/prj.conf
```
mniestroj, line 126: That should not be needed anymore, at least on `main`.

mniestroj (commented):
Looks great already!

Most important thing: can we move `hardware-map.yaml` to runner?

Another (but can be part of another PR) thing is removing all the "dfu" specifics and make it more generic to run all samples, not just DFU.

.github/workflows/twister_dfu_nrf52840dk.yml
szczys: The OrangePis used for self-hosted runners are not easy to access. For that reason it is better to make all configuration files exist as GitHub Secrets. I have an improvement in mind that I will outline in a design document during next sprint. It will outline a process where new hardware can be added to a runner without any local configuration. I recommend going with what we have here for this PR and then refining as we review that design doc.

.github/workflows/twister_dfu_nrf52840dk.yml
szczys: In recent discussion with @sam-golioth, we'd like to have cycle testing just for DFU. I'd like to merge this to focus only on the DFU sample for now and see how it performs is scheduled (an manual) runs over the next sprint, then open up a workflow that runs all samples.

.github/workflows/twister_dfu_nrf52840dk.yml
mniestroj: Well, hardware configuration depends on what is connected to the runner. This is something that needs to be checked manually, e.g. by serial connection with Linux or with SSH. And the only way you know what is connected is by checking it physically (in order to know which DevKit/board is connected). Another thing is `/dev/serial/*` node to be used. Last thing is actually the list of connected boards (there might be many of them connected to a single runner).

I really don't see why we need to store "hardware map" (information about what is connected to a single runner) in a central database (which is GitHub Actions Secrets).

szczys: It's a pain to get into the shell of these devices. They are running on their own network, so you need to connect to that network and then ssh into them. At that point the person who is running it needs to know enough about how they're set up to find the device serial endpoint and then know enough about formatting the hardware map files to create one. Then run a test using that file. It's a lot of overhead for people who don't work with these systems often. I've been working with these self-hosted runners for quite a while now and it is still a pain for me to do this.

My idea is to have a workflow that will get the output of `ls -la /dev/serial/by-id` which can then be used to create the hardware map. This way, adding new hardware means just plugging it in and running the workflow.

I think we should continue this discussion once I have submitted the design document for this.

szczys:
I added three commits to this PR that fix two issues that were causing this workflow to fail:

1. added `golioth rollback '*'` command which will rollback any existing releases before starting Twister
2. workflow adds a delay to the top of main() so logs are not missed during test.

The second point is not a permanent fix, but it will be enough for us to reliably run this test while a permanent fix is found.

.github/workflows/twister_dfu_nrf52840dk.yml
```diff
@@ -125,6 +125,8 @@ jobs:
           PATH=$PATH:/opt/nrf-command-line-tools/bin
           nrfjprog -i
           golioth releases rollback '*'
+          # FIXME: workaround when missing the initial IDLE log
+          sed -i '/LOG_DBG("Start DFU sample");/a \\tk_sleep(K_SECONDS(10));' modules/lib/golioth/samples/dfu/src/main.c
```
mniestroj, line 129: How about adding `CONFIG_BOOT_DELAY=10000`? Would that help as well? We could put that into `sample.yaml` if it would work.

.github/workflows/twister_dfu_nrf52840dk.yml
sam-golioth: Once we move on running all samples, we'll want each test (i.e. each sample) to show up separately in the results in Github CI, so depending on how Twister/Github spit out the results, we may want to run them all separately anyway.

.github/workflows/twister_dfu_nrf52840dk.yml
```diff
@@ -0,0 +1,139 @@
...
+    # Run workflow at the start of every day (12 AM UTC)
+    - cron: "0 0 * * *"
+
+  workflow_dispatch:
+
+jobs:
+  # Assumptions made about the self-hosted runner:
+  #
+  #  1. Has installed the GitHub Actions self-hosted runner service
+  #  2. Has nrfjprog already installed at /opt/nrf-command-line-tools
```
sam-golioth, line 16: Just a note for posterity, we should look into options for making this managed automatically (docker, conda, etc.). As we grow the number of runners and tests, managing these dependencies and making sure they're all on the same version and such will get unwieldy.

.github/workflows/twister_dfu_nrf52840dk.yml
```diff
@@ -124,6 +124,7 @@ jobs:
           source env_dfu_nrf52840dk.sh
           PATH=$PATH:/opt/nrf-command-line-tools/bin
           nrfjprog -i
+          golioth releases rollback '*'
```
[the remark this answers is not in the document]
szczys: I thought about that, and two things drove this decision:

* removing releases is destructive (you can't get the deleted binaries back),  but rollback can be undone.
* I think if v2.0.0 exists, we probably want the test to fail because it indicates some previous trouble that we should pay attention to

If you feel strongly about deleting instead of rolling back, let me know and I'll change it.

.github/workflows/twister_dfu_nrf52840dk.yml
szczys: Yes! I didn't know about this symbol, it's a much better way to handle it. Thanks.

.github/workflows/twister_dfu_nrf52840dk.yml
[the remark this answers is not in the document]
mniestroj: There should be no other than 2.0.0 releases/artifacts, as pytest should be the only user of this project/account. Right? So even when `golioth releases rollback '*'` will disable DFU of the device, the test will still fail due to existing artifacts/releases. So the only difference it makes is that the test will fail in a different way, but it **will** fail. This is why I would either do artifacts/releases removal **or** nothing at all. Does it make sense?

mniestroj (approved):
Thanks a lot for that part! This is a big milestone, which was on my TODO list for over 2 years ;) Finally there!

PR #371, opened 2023-04-21
samples: dfu: README: use $(west topdir) instead of WEST_ROOT

mniestroj:
By referencing `$(west topdir)` the whole command can be simply copy-pasted
by user and it will work as is, instead of requiring to manually replace
`WEST_ROOT` with the directory where west was initialized.


szczys (approved):
Wow, neat trick! LGTM.

PR #372, opened 2023-04-21
README: reference nRF9160 DK instead of nRF9160 Feather

mniestroj:
Since DK is the continuously verified board instead Feather, reference the
former in README file.


szczys (approved):
LGTM!

PR #373, opened 2023-04-21
samples: logging: pytest: cleanup import modules list

mniestroj:
Some of the imported modules were not used. Remove those.


szczys (approved):
LGTM!

PR #374, opened 2023-04-21
samples: dfu: disable settings in system client for vanilla Zephyr tests

mniestroj:
Disable settings support in system client, so that PSK-ID and PSK will be
pulled in from environment variables and baked into the firmware. This
allows to omit credentials provisioning.

This is a follow-up of commit 6f1e59270834 ("samples: dfu: disable settings
in system client for NCS runtime test").

### DIFF
--- a/samples/dfu/sample.yaml
+++ b/samples/dfu/sample.yaml
@@ -15,6 +15,8 @@ tests:
     platform_allow: >
       mimxrt1060_evkb
       nrf52840dk_nrf52840
+    extra_args:
+      dfu_CONFIG_GOLIOTH_SYSTEM_SETTINGS=n
   sample.golioth.dfu.ncs:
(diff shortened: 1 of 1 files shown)


szczys (approved):
I'd just like to confirm, this change only affects twister. User running the following command will still see the value in prj.con used:

```
west build -b nrf52840dk_nrf5840 . -p --sysbuild
```

Is that correct?

mniestroj:
> I'd just like to confirm, this change only affects twister. User running the following command will still see the value in prj.con used:
> 
> ```
> west build -b nrf52840dk_nrf5840 . -p --sysbuild
> ```
> 
> Is that correct?

Yes, that is correct. The value added in twister is applied automatically by twister (by appending `-DOVERLAY_CONFIG=` to the command line invocation before building, I think).

PR #375, opened 2023-04-24
samples: dfu: pytest: convert os.path.* to pathlib

mniestroj:
Convert all `os.path.*` invocations to `pathlib`, in order to improve code
readability and consistency.


szczys (approved):
I tested this on the nRF52840 and it works for me!

PR #379, opened 2023-05-02
.github: test_nrf52840dk use subdirectory

szczys:
When running workflows on self-hosted runners, checkout and working-directory should use a subdirectory so that different workflows don't collide.

Generally, this means Twister-based workflows can use the same subdirectory to share common files. For example, modules generally don't differ from one test to the next, and if there is a different, the `west update` step will correct for it.

However, tests that don't use Twister (specifically, the ones that don't use a Docker container) need different user permissions on the files. Using subdirectories keeps these separate while still minimizing the amount of cloned data on subsequent runs.


szczys:
Sorry for all the noise on this PR. I saw that overnight the nRF9160dk test failed for the same reason so I added a commit to make the changes there as well.

I had a few false-starts in how to reference a previous commit in a subsequent commit message so checkpatch set me straight. (I'm unsure of why my local checkpatch didn't catch the issue?)

This is now ready for review. Thanks!

.github/workflows/test_nrf9160dk.yml
```diff
@@ -94,15 +94,21 @@ jobs:
...
         uses: actions/checkout@v3
+        with:
+          path: connection_tests
 
       - name: Download build tarball
         uses: actions/download-artifact@v3
         with:
           name: build.tar.gz
-          path: samples/test
+          path: connection_tests/samples/test
```
mniestroj, line 111: Change is fine, but referencing commit by sha1 won't work after this PR will be merged using GitHub, as GitHub does automatic rebase and sha1 of the referenced commit will change. So maybe just drop this sha1 and leave commit title only.

mniestroj (commented):
LGTM.

Just a minor comment about referencing commit by sha1.

PR #382, opened 2023-05-22
samples: move parts of prj.conf to Kconfig.defconfig

mniestroj:
The general idea is to reduce duplication across different samples' `prj.conf` files and configure
common settings using Kconfig default values.


szczys:
I think moving these configurations out of the prj.conf makes them less visible to the user. When I'm working on adding a feature to an existing application I often look at the prj.conf in our sample files as a clue to what Kconfig symbols may also be needed.

For the main process thread it's not too hard to figure out for yourself. But for the mbedtls options I think it's quite helpful.

mniestroj:
> I think moving these configurations out of the prj.conf makes them less visible to the user. When I'm working on adding a feature to an existing application I often look at the prj.conf in our sample files as a clue to what Kconfig symbols may also be needed.
> 
> For the main process thread it's not too hard to figure out for yourself. But for the mbedtls options I think it's quite helpful.

That is a good point. This is exactly the reason why I haven't done that before. The thing is... we are trying to support both vanilla Zephyr and NCS at the same time. This is almost impossible to do with `prj.conf` approach, since almost all mbedTLS configuration options are different in Zephyr and different in NCS (they use their own fork of mbedTLS, with their Kconfig files).

So during the work on NCS and cert-based auth, I figured that moving some options from `prj.conf` to `Kconfig.defconfig` would help cleanup the dependencies. The alternative would be to move from `prj.conf` to `boards/BOARD.conf` files, which would increase duplication (instead of decreasing it), probably making it harder to maintain in the long term.

See https://github.com/golioth/golioth-zephyr-sdk/pull/383 which is built on top of this PR. This adds (or fixes) support for cert-based auth in NCS.

TLDR; I'd prefer not to do this if I would have better/alternative solution. Any ideas are welcome.

mniestroj:
Actually the alternative approach would be moving from `prj.conf` to snippets, as in https://github.com/golioth/golioth-zephyr-sdk/pull/384. However, this is not supported yet in NCS (maybe in next release it will) and there are issues with applying `samples.yaml` configuration overrides, which would break `twister` builds/tests.

sam-golioth:
> Actually the alternative approach would be moving from `prj.conf` to snippets, as in #384. However, this is not supported yet in NCS (maybe in next release it will) and there are issues with applying `samples.yaml` configuration overrides, which would break `twister` builds/tests.

Why do snippets work for the changes in #384 but not the changes in this PR?

mniestroj:
> > Actually the alternative approach would be moving from `prj.conf` to snippets, as in #384. However, this is not supported yet in NCS (maybe in next release it will) and there are issues with applying `samples.yaml` configuration overrides, which would break `twister` builds/tests.
> 
> Why do snippets work for the changes in #384 but not the changes in this PR?

Zephyr snippets is a new feature and there are issues, e.g. with twister, as seen in #384 github actions (they fail). But overall, snippets approach could be used in this PR as well. The thing is... it's not ready yet for using, so my proposal in this PR was to move to `Kconfig.defconfig` and once snippets get more adoption and some issues are fixed, we could move to snippets for better clarity (than with `Kconfig` defaults).

sam-golioth (approved):

PR #389, opened 2023-06-02
samples: prepare for NCS bump to 2.4.0

mniestroj:
There are some minor issues when trying to bump NCS to 2.4.0:
 * `CONFIG_NET_SHELL=y` results in build failure, so this PR includes a commit that sets this option
   via Kconfig, instead of `prj.conf` (this also deduplicates lots of configuration),
 * `void main()` functions result in build warnings, so this PR includes a conversion to `int
   main()`, which is what upstream Zephyr switched to,
 * ~~`CONFIG_LTE_AUTO_INIT_AND_CONNECT=y` got deprecated in NCS 2.4.0, so there is a commit to stop
   using it; this also improves consistency with WiFi mgmt APIs, which are called by
   `samples/common/` code as well.~~


szczys (approved):
LGTM!

PR #393, opened 2023-06-13
bump NCS to 2.4.0

mniestroj:
Newest NCS version is desired to pull zcbor API changes.

Satisfy build-time check of `MBEDTLS_PK_C`, which requires either ECP or RSA
to be enabled.

Select RSA for cert-based auth, as it is no longer selected by default.

Use `CONFIG_LTE_CONNECTIVITY` instead of `CONFIG_LTE_AUTO_INIT_AND_CONNECT`, as
the latter is deprecated. Select IPv6 (with certains options disabled), as
it is a requirement of `CONFIG_LTE_CONNECTIVITY`.

NCS' west manifest now includes `qcbor` module. Rename our `qcbor` to
`golioth-qcbor`, so that name does not conflict.

Fixes: https://github.com/golioth/firmware-issue-tracker/issues/141


sam-golioth:
> Select IPv6 (with certains options disabled)

Were we using IPv6 before? What's the RAM/ROM impact of enabling this?

sam-golioth (approved):

mniestroj:
> > Select IPv6 (with certains options disabled)
> 
> Were we using IPv6 before? What's the RAM/ROM impact of enabling this?

We didn't use and still don't use IPv6. Enabling it is required to satisfy NCS broken Kconfig dependencies. I haven't checked what is the RAM/ROM impact, as there is no way to keep IPv6 disabled with new NCS version (so can't really isolate how IPv6 alone affects application footprint).

PR #394, opened 2023-06-13
kconfig: add options for maximum PSK-ID and PSK lengths

mniestroj:
PSK-ID length was always hardcoded to maximum 64 characters. As a result
long project names with long PSK-IDs (due to
`long-device-id@long-project-name` PSK-ID format) would not be supported.

PSK had the same issue, but only when used with NCS.

Introduce options for both, PSK-ID and PSK, so that long credentials can be
supported without changing SDK codebase.

Alternative to: #359

### DIFF
--- a/net/golioth/Kconfig
+++ b/net/golioth/Kconfig
@@ -310,6 +310,23 @@ config GOLIOTH_SYSTEM_SETTINGS
 	  When selected, Golioth credentials will be loaded from settings
 	  subsystem.
 
+if GOLIOTH_SYSTEM_SETTINGS
+
+config GOLIOTH_SYSTEM_CLIENT_PSK_ID_MAX_LEN
+	int "Max length of PSK ID"
+	default 64
+	help
+	  Maximum length of PSK ID, in bytes.
+
+config GOLIOTH_SYSTEM_CLIENT_PSK_MAX_LEN
+	int "Max length of PSK"
+	default MBEDTLS_PSK_MAX_LEN if MBEDTLS_BUILTIN
+	default 64
+	help
+	  Maximum length of PSK, in bytes.
+
+endif # GOLIOTH_SYSTEM_SETTINGS
+
 endif # GOLIOTH_SYSTEM_CLIENT
 
 endif # GOLIOTH
--- a/net/golioth/system_client.c
+++ b/net/golioth/system_client.c
@@ -475,9 +475,9 @@ void golioth_system_client_stop(void)
  * credentials are stored. This means that we need to allocate memory for
  * credentials ourselves.
  */
-static uint8_t golioth_dtls_psk[CONFIG_GOLIOTH_SYSTEM_CLIENT_PSK_MAX_LEN];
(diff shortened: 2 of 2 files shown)


mniestroj:
FYI: @danyocom

sam-golioth:
@ChrisGammell ran into the same problem and it took him a while to figure out what the issue was. It looks like it wasn't immediately obvious to @danyocom either. @mniestroj can we also add an error message to the console if the length exceeds the configured value? That way it should be pretty straightforward for users who encounter this issue to determine what caused it and how to fix it.

szczys (dismissed):
LGTM!

net/golioth/Kconfig
```diff
@@ -323,6 +323,23 @@ config GOLIOTH_SYSTEM_SETTINGS
 	  When selected, Golioth credentials will be loaded from settings
 	  subsystem.
 
+if GOLIOTH_SYSTEM_SETTINGS
```
szczys, line 326: I like that there's an ASSERT to detect values that are too long:
```
/home/mike/golioth-zephyr-workspace/modules/lib/golioth/net/golioth/system_client.c:38:1: error: static assertion failed: "PSK exceeds mbedTLS configured maximum PSK length"
   38 | BUILD_ASSERT(sizeof(TLS_PSK) - 1 <= CONFIG_MBEDTLS_PSK_MAX_LEN,
      | ^~~~~~~~~~~~
```

However, selecting `GOLIOTH_SYSTEM_SETTINGS` means compiled-in credentials can't access the new length settings.

```
config GOLIOTH_SYSTEM_CLIENT_PSK_ID
	string "PSK ID"
	depends on !GOLIOTH_SYSTEM_SETTINGS
```

szczys (commented):
When testing `samples/settings` on esp32 I am unable to store setting;

```
uart:~$ settings set golioth/psk 12345
Setting golioth/psk to 12345
Failed to save setting golioth/psk:12345
```

szczys:
> @mniestroj can we also add an error message to the console if the length exceeds the configured value? That way it should be pretty straightforward for users who encounter this issue to determine what caused it and how to fix it.

I agree with this. I tested a psk-id that was 148 chars long with the limit set to 128. The credential save succeeded without error messages but the device failed to connect.

net/golioth/Kconfig
```diff
@@ -323,6 +323,23 @@ config GOLIOTH_SYSTEM_SETTINGS
...
+
+config GOLIOTH_SYSTEM_CLIENT_PSK_ID_MAX_LEN
+	int "Max length of PSK ID"
+	default 64
+	help
+	  Maximum length of PSK ID, in bytes.
+
+config GOLIOTH_SYSTEM_CLIENT_PSK_MAX_LEN
```
szczys, line 334: I am unable to get this to work when my PSK is longer than 32 characters.

sam-golioth (changes requested):
Looks good once we add the error message for exceeding the configured length

sam-golioth (approved):

PR #395, opened 2023-06-13
treewide: migrate from QCBOR to zcbor

mniestroj:
zcbor is supported by Zephyr upstream, so there is no need to maintain the glue layer and pulling
(security) fixes. Additionally zcbor is more lightweight and fits even better into MCU based
projects with little resources.

Depends on: #399


include/net/golioth/rpc.h
```diff
@@ -11,11 +11,12 @@
 #include <zephyr/net/coap.h>
 #include <zephyr/kernel.h>
 
-struct _QCBOREncodeContext;
-struct _QCBOREncodeContext;
-
-typedef struct _QCBORDecodeContext QCBORDecodeContext;
-typedef struct _QCBOREncodeContext QCBOREncodeContext;
+#ifdef CONFIG_ZCBOR
```
sam-golioth, line 14: Is this necessary?` CONFIG_ZCBOR` is a hard requirement for the RPC service, I think it would be cleaner to make sure that `rpc.h` isn't included by any other source files if `CONFIG_ZCBOR` (or more likely `CONFIG_GOLIOTH_RPC`) isn't selected. Alternatively, I'd also be open to making `CONFIG_ZCBOR` selected by `CONFIG_GOLIOTH` instead of by the individual services. I can't imagine anybody using Golioth but not using fw update, rpc, and settings. It's not as direct, but it would simplify things for us.

logging/log_backend_golioth.c
```diff
@@ -26,8 +25,7 @@
 #define CBOR_SPACE_RESERVED	8
 
 struct golioth_cbor_ctx {
-	QCBOREncodeContext encode_ctx;
-	UsefulBuf encode_bufc;
+	zcbor_state_t zse[1 /* num_backups */ + 2];
```
sam-golioth, line 28: NIT: Let's use a preprocessor constant here for the magic number.

net/golioth/rpc.c
```diff
@@ -118,38 +153,49 @@ static int on_rpc(struct golioth_req_rsp *rsp)
...
+		if (!ok) {
+			LOG_ERR("Failed to encode RPC '%s'", "detail");
+			return -ENOMEM;
+		}
 
-	k_mutex_unlock(&client->rpc.mutex);
+		ok = zcbor_map_start_encode(zse, SIZE_MAX);
+		if (!ok) {
+			LOG_ERR("Did not start CBOR map correctly");
+			return -ENOMEM;
```
sam-golioth, line 165: All of these returns leave `client->rpc.mutex` locked

sam-golioth (changes requested):
Looks good! Just need to fix the mutex leak.

include/net/golioth/rpc.h
sam-golioth: > I'd prefer to keep net/golioth.h minimal, with just "core" code. Things like RPC, LightDB State, LightDB Stream, etc. should go into separate headers.

I agree with this, and I think if we had it that way then we wouldn't need this guard (applications that include `rpc.h` but don't enable `CONFIG_GOLIOTH_RPC` will rightfully fail to compile). Those changes are out of scope for this ticket though, so I'm good with what's here.

sam-golioth (approved):

PR #398, opened 2023-06-15
Move hardcoded certificates from Golioth subsystem into samples

sam-golioth:
This PR removes the management of TLS certificates from the Golioth library. Clients are now responsible for setting TLS certificate credentials in the TLS Credential subsystem before starting the Golioth system, as well as persisting them on the device. In order to facilitate prototyping and experimentation, a new `hardcoded_credentials` module has been added to `samples/common` to provide the same functionality that used to be baked directly into the SDK. This gives application developers the flexibility to implement their own provisioning and storage mechanisms, while maintaining the same low-effort experience for users of the samples.

Note that this doesn't actually provide any facility for provisioning certificates, it only removes the parts of the SDK that prevented a customer from implementing provisioning themselves. In a future PR, I'll add a sample that shows one way a developer could provision certificates onto the device. This PR was getting larger than I initially expected, so I decided to break it into parts.

Note too that the new `hardcoded_credentials` module is also where I plan to put hardcoded PSK credentials for the samples after we remove that from the SDK. Bringing the hardcoded credentials code up into the samples should make it more obvious to users when they need to set credentials in Kconfig vs provisioned, as well as make the SDK itself cleaner.


net/golioth/CMakeLists.txt
```diff
@@ -12,23 +12,20 @@ zephyr_library_sources_ifdef(CONFIG_GOLIOTH_SETTINGS settings.c)
 zephyr_library_sources_ifdef(CONFIG_GOLIOTH_SYSTEM_CLIENT system_client.c)
 
 if(CONFIG_GOLIOTH_AUTH_METHOD_CERT)
-  foreach(type IN ITEMS ca crt key)
-    string(TOUPPER ${type} type_upper)
-    set(path ${CONFIG_GOLIOTH_SYSTEM_CLIENT_${type_upper}_PATH})
+  set(path ${CONFIG_GOLIOTH_SYSTEM_CLIENT_CA_PATH})
```
sam-golioth, line 15: The CA cert stays hardcoded in the Golioth library for now because it is the same across the entire fleet. Eventually, we'll need a way to provision this certificate, but that is part of a larger story around rolling the CA certificate that is out of scope of these changes.

net/golioth/system_client.c
```diff
@@ -454,11 +452,19 @@ K_THREAD_DEFINE(golioth_system, CONFIG_GOLIOTH_SYSTEM_CLIENT_STACK_SIZE,
...
 	if (IS_ENABLED(CONFIG_GOLIOTH_AUTH_METHOD_PSK)) {
 		golioth_settings_check_credentials();
+	} else if (IS_ENABLED(CONFIG_GOLIOTH_AUTH_METHOD_CERT)) {
+		ret = golioth_check_cert_credentials();
+	}
+
+	if (ret == 0) {
+		k_sem_give(&sys_client_started);
+	} else {
+		LOG_WRN("Error loading TLS credentials, golioth system client was not started");
```
mniestroj, line 465:
```suggestion
	int err = 0;

	if (IS_ENABLED(CONFIG_GOLIOTH_AUTH_METHOD_PSK)) {
		golioth_settings_check_credentials();
	} else if (IS_ENABLED(CONFIG_GOLIOTH_AUTH_METHOD_CERT)) {
		err = golioth_check_cert_credentials();
	}

	if (err) {
		LOG_WRN("Error loading TLS credentials, golioth system client was not started");
	} else {
		k_sem_give(&sys_client_started);
```

net/golioth/system_client.c
```diff
@@ -159,12 +147,38 @@ static int golioth_check_credentials(const uint8_t *psk_id, size_t psk_id_len,
...
 }
 
+static int golioth_check_cert_credentials(void)
+{
+	size_t cred_len = 0;
+	int err = tls_credential_get(CONFIG_GOLIOTH_SYSTEM_CLIENT_CREDENTIALS_TAG,
+				     TLS_CREDENTIAL_SERVER_CERTIFICATE, NULL, &cred_len);
+	if (-ENOENT == err) {
+		LOG_WRN("Certificate authentication configured, but no client certificate found");
+		goto finish;
```
mniestroj, line 157:
```suggestion
		return err;
```

net/golioth/system_client.c
```diff
@@ -159,12 +147,38 @@ static int golioth_check_credentials(const uint8_t *psk_id, size_t psk_id_len,
...
+	}
+
+	err = tls_credential_get(CONFIG_GOLIOTH_SYSTEM_CLIENT_CREDENTIALS_TAG,
+				 TLS_CREDENTIAL_PRIVATE_KEY, NULL, &cred_len);
+	if (-ENOENT == err) {
+		LOG_WRN("Certificate authentication configured, but no private key found");
+		goto finish;
```
mniestroj, line 164:
```suggestion
		return err;
```

net/golioth/system_client.c
```diff
@@ -159,12 +147,38 @@ static int golioth_check_credentials(const uint8_t *psk_id, size_t psk_id_len,
 	return err;
 }
 
+static int golioth_check_cert_credentials(void)
+{
+	size_t cred_len = 0;
+	int err = tls_credential_get(CONFIG_GOLIOTH_SYSTEM_CLIENT_CREDENTIALS_TAG,
+				     TLS_CREDENTIAL_SERVER_CERTIFICATE, NULL, &cred_len);
+	if (-ENOENT == err) {
```
mniestroj, line 155:
```suggestion
	if (err == -ENOENT) {
```
And how about other error codes?

net/golioth/system_client.c
```diff
@@ -159,12 +147,38 @@ static int golioth_check_credentials(const uint8_t *psk_id, size_t psk_id_len,
...
+		LOG_WRN("Certificate authentication configured, but no client certificate found");
+		goto finish;
+	}
+
+	err = tls_credential_get(CONFIG_GOLIOTH_SYSTEM_CLIENT_CREDENTIALS_TAG,
+				 TLS_CREDENTIAL_PRIVATE_KEY, NULL, &cred_len);
+	if (-ENOENT == err) {
```
mniestroj, line 162:
```suggestion
	if (err == -ENOENT) {
```

net/golioth/system_client.c
```diff
@@ -159,12 +147,38 @@ static int golioth_check_credentials(const uint8_t *psk_id, size_t psk_id_len,
...
+		LOG_WRN("Certificate authentication configured, but no private key found");
+		goto finish;
+	}
+
+	if (-EFBIG == err) {
```
mniestroj, line 167:
```suggestion
	if (err == -EFBIG) {
```

mniestroj (dismissed):
I like the direction this is going to, i.e. decoupling "system client" from TLS credentials.

We need something similar for PSK auth as well, so that user (or in our case `samples/` implementation) will be responsible for setting credentials.

`s/libray/library/` typo in commit message.

It looks like those 2 commits are not bisectable, so probably they should be squashed.

net/golioth/system_client.c
sam-golioth: I prefer to have a single return point from each function. In a small function like this, it doesn't make much difference, but in larger functions it helps to make the code more readable (subjective) and reduces the likelihood of bugs arising from improper cleanup (objective).

net/golioth/system_client.c
sam-golioth: This was put in specifically to guard against https://github.com/pion/dtls/issues/529 (spent about 30 minutes confused before I remembered that one). I don't think there's value in checking other error codes. Right now, the only other one that is specified in the API is `-EACCESS`. This isn't actually returned anywhere in the implementation, but I imagine it's there in the case that a future TLS Credentials subsystem implementation uses a secure storage backend, which could allow writing keys but not reading them, or only allow reading keys from secure code.

net/golioth/system_client.c
sam-golioth: I prefer to have the happy path first, and error handling in the `else` clause. I'll update to the `err` naming to match the rest of the file.

sam-golioth:
> We need something similar for PSK auth as well, so that user (or in our case `samples/` implementation) will be responsible for setting credentials.

Yup. In the PR description I talk about using the same `hardcoded_credentials` module to handle PSKs. It will be done as part of https://github.com/golioth/firmware-issue-tracker/issues/145

net/golioth/CMakeLists.txt
```diff
@@ -12,23 +12,20 @@ zephyr_library_sources_ifdef(CONFIG_GOLIOTH_SETTINGS settings.c)
...
+    else()
+      message(FATAL_ERROR "${path} configured by CONFIG_GOLIOTH_SYSTEM_CLIENT_CA_PATH does not exist")
     endif()
+  endif()
 
-    generate_inc_file_for_target(app ${path}
-      ${ZEPHYR_BINARY_DIR}/include/generated/golioth-systemclient-${type}.inc)
-  endforeach()
+  generate_inc_file_for_target(app ${path}
+    ${ZEPHYR_BINARY_DIR}/include/generated/golioth-systemclient-${type}.inc)
```
szczys, line 28:
```suggestion
    ${ZEPHYR_BINARY_DIR}/include/generated/golioth-systemclient-ca.inc)
```

I don't think type is being set now that the `foreach` command is removed. The result is that the generated file is missing the `ca` in its title:

```
golioth-systemclient-.inc
golioth-systemclient-crt.inc
golioth-systemclient-key.inc
```

Which results in a build error:
```
/home/mike/golioth-zephyr-workspace/modules/lib/golioth/net/golioth/system_client.c:50:10: fatal error: golioth-systemclient-ca.inc: No such file or directory
   50 | #include "golioth-systemclient-ca.inc"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

szczys (changes requested):

net/golioth/CMakeLists.txt
sam-golioth: Thanks! I had been trying to reproduce this locally with no luck. Turns out I had a stale cached version of `prj.conf` in my editor that was displaying `CONFIG_GOLIOTH_AUTH_METHOD_CERT=y` without that actually be saved to disk 😞.

szczys (approved):
LGTM. I tested this on an ESP32 and it works great!

PR #400, opened 2023-06-21
feat: move to using Golioth HIL images

hasheddan:
Updates to use the Golioth HIL images from Google Artifact Registry.


sam-golioth:
This PR has gone stale, gonna close for now and revisit later as we enable HIL tests in the CP SDK

PR #403, opened 2023-06-27
samples: add certificate provisioning sample

sam-golioth:
Adds an example of one way to provision certificates. Certificate and private key are loaded onto the device's filesystem using mcumgr. At bootup, the sample app reads the certificate and key from the filesystem and then loads them into the Zephyr TLS credential store.

Still to come (in future PRs):
- Refactor of key generation script and moving it into `samples/common` for use by any sample.
- Automated testing


szczys (approved):
LGTM. Tested on ESP32 and found to be working. Good stuff!

(Note: I used newtmgr instead of mcumgr and had no issues with all of the same commands/params)

sam-golioth:
> (Note: I used newtmgr instead of mcumgr and had no issues with all of the same commands/params)

Nice! That makes sense, since mcumgr is part of the Mynewt project. Do you prefer newtmgr over mcumgr?

PR #404, opened 2023-06-28
samples: refactor certificate generation script

sam-golioth:
- Separate commands for root and device certificates

- Don't require device to already exist in Golioth console

- Move to samples/common/scripts


mniestroj:
> - Separate commands for root and device certificates
> - Don't require device to already exist in Golioth console
> - Move to samples/common/scripts

Can we split those 3 changes to separate commits to make is easier to review (and get back to it once needed with `git log`)?

sam-golioth:
> > * Separate commands for root and device certificates
> > * Don't require device to already exist in Golioth console
> > * Move to samples/common/scripts
> 
> Can we split those 3 changes to separate commits to make is easier to review (and get back to it once needed with `git log`)?

I split it into 2 commits, one for moving the file and another for the code changes, which should make the diff easier to use.

samples/certificate_provisioning/src/main.c
```diff
@@ -0,0 +1,145 @@
...
+	if (err < 0) {
+		LOG_ERR("Could not open %s", path);
+		goto finish;
+	}
+
+	/* NOTE: cred_buf is used directly by the TLS Credentials library, and so must remain
+	 * allocated for the life of the program.
+	 */
+
+	void *cred_buf = malloc(dirent.size);
```
mniestroj, line 77: Please put variable declarations at the top of function, so it is easy to see which (and how many) variables are used within function scope.

samples/certificate_provisioning/src/main.c
```diff
@@ -0,0 +1,145 @@
...
+
+	if (cred_buf == NULL) {
+		LOG_ERR("Could not allocate space to read credential");
+		err = -ENOMEM;
+		goto finish_with_file;
+	}
+
+	err = fs_read(&file, cred_buf, dirent.size);
+
+	if (err < 0) {
```
mniestroj, line 87: Please remove newline between those lines, as the result is checked right away. This will make it consistent with Zephyr coding style.

samples/hello/scripts/generate-keys.py
```diff
@@ -52,12 +51,29 @@ async def main(config_path, api_key, device_name):
     # Create and self-sign a corresponding public key / certificate
     run(f'openssl req -x509 -new -nodes -key "{root_ca_key}" -sha256 -subj "/C=US/CN=Root {ROOT_CA_NAME}" -days 1024 -out "{root_ca_crt}"')
 
+    if upload:
+        # Add (upload) generated CA certificate to Golioth
+        with root_ca_crt.open('rb') as fp:
+            await project.certificates.add(fp.read(), cert_type='root')
+
+@click.command()
+@click.option('--device-id', help="Primary Device Hardware ID (e.g. a serial number)",
```
mniestroj: "device-id" and "hardware-id" are probably two separate things in Golioth nomenclature. We should use one name for one thing. E.g. when you list devices with `goliothctl device list` then there is `id:"xxxxxxxxxxxxxxxx"` field shown and then `hardware_ids:"xxxxx"`.

samples/hello/scripts/generate-keys.py
```diff
@@ -6,43 +6,42 @@
...
-    client_name = f"{project.id}-{device.name}"
+        if project_name:
+            project = await client.project_by_name(project_name)
+        else:
+            project = await client.default_project()
 
-    root_ca_crt = SCRIPTS_DIR / f'{ROOT_CA_NAME}-{project.id}.crt.pem'
-    root_ca_key = SCRIPTS_DIR / f'{ROOT_CA_NAME}-{project.id}.key.pem'
+    root_ca_crt = output / f'{ROOT_CA_NAME}-{project_name}.crt.pem'
+    root_ca_key = output / f'{ROOT_CA_NAME}-{project_name}.key.pem'
```
mniestroj: In the first function there is `project_name` and in the second there is `project_id`. Also, in the second function argument description there is "Project slug". It is all very confusing.

samples/hello/scripts/generate-keys.py
```diff
@@ -52,12 +51,29 @@ async def main(config_path, api_key, device_name):
...
+@click.option('--device-id', help="Primary Device Hardware ID (e.g. a serial number)",
+              required=True)
+@click.option('--project-id', help="Project slug", required=True)
+@click.option('--root-ca-crt', type=Path, help="Root certificate that will sign device certificate",
+              required=True)
+@click.option('--root-ca-key', type=Path, help="Root private key that will sign device certificate",
+              required=True)
+@click.option('-o', '--output', type=Path, default=Path('.'), help="Output directory")
+async def device(device_id, project_id, root_ca_crt, root_ca_key, output):
+    client_name = output / f"{project_id}-{device_id}"
```
mniestroj: `client_name` is a path, so this variable name is confusing.

samples/hello/scripts/generate-keys.py
```diff
@@ -52,12 +51,29 @@ async def main(config_path, api_key, device_name):
...
+
+@click.command()
+@click.option('--device-id', help="Primary Device Hardware ID (e.g. a serial number)",
+              required=True)
+@click.option('--project-id', help="Project slug", required=True)
+@click.option('--root-ca-crt', type=Path, help="Root certificate that will sign device certificate",
+              required=True)
+@click.option('--root-ca-key', type=Path, help="Root private key that will sign device certificate",
+              required=True)
+@click.option('-o', '--output', type=Path, default=Path('.'), help="Output directory")
```
mniestroj: It's great that `-o` `--output` option is added, but please such changes separate for easier review. There is a bit too much happening at once, which is hard to track during review.

samples/certificate_provisioning/src/main.c
```diff
@@ -0,0 +1,145 @@
...
+#define STORAGE_PARTITION_LABEL	storage_partition
+#define STORAGE_PARTITION_ID	FIXED_PARTITION_ID(STORAGE_PARTITION_LABEL)
+
+FS_LITTLEFS_DECLARE_DEFAULT_CONFIG(cstorage);
+static struct fs_mount_t littlefs_mnt = {
+	.type = FS_LITTLEFS,
+	.fs_data = &cstorage,
+	.storage_dev = (void *)STORAGE_PARTITION_ID,
+	.mnt_point = "/lfs1"
+};
```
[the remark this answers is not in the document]
sam-golioth: This commit was included by mistake, it's from an older PR. I'll open another PR to address these comments.

samples/common/scripts/generate-keys.py
```diff
@@ -0,0 +1,94 @@
...
+
+    if upload:
+        # Add (upload) generated CA certificate to Golioth
+        with root_ca_crt.open('rb') as fp:
+            await project.certificates.add(fp.read(), cert_type='root')
+
+@click.command()
+@click.option('--device-id', help="Primary Device Hardware ID (e.g. a serial number)",
+              required=True)
+@click.option('--project-id', help="Project slug", required=True)
```
mniestroj, line 62:
```suggestion
@click.option('--project-id', help="Project ID", required=True)
```

samples/common/scripts/generate-keys.py
```diff
@@ -0,0 +1,94 @@
...
+    run(f'openssl req -x509 -new -nodes -key "{root_ca_key}" -sha256 -subj "/C=US/CN=Root {ROOT_CA_NAME}" -days 1024 -out "{root_ca_crt}"')
+
+    if upload:
+        # Add (upload) generated CA certificate to Golioth
+        with root_ca_crt.open('rb') as fp:
+            await project.certificates.add(fp.read(), cert_type='root')
+
+@click.command()
+@click.option('--device-id', help="Primary Device Hardware ID (e.g. a serial number)",
+              required=True)
```
mniestroj, line 61: In this line there are 3 things mentioned (device id, primary hardware id, serial number) and it is hard to figure out is really requested from the user. Also `(e.g. a serial number)` suggests that this is **something** that user can select (which to some extent is true, but probably not at this point anymore). I would prefer to keep using "primary hardware id" or "hardware id".
```suggestion
@click.option('--primary-hardware-id', help="Primary Device Hardware ID,
              required=True)
```

sam-golioth: So what this actually will map to in the console is the "Device Name". This goes into the `CN` field in the certificate, which directly populates the Device Name field in the console. The device name is combined with a timestamp to create the "Primary Hardware ID" in the console. It definitely is something that a user can select, and I think a serial number is the best name for most people. I'll update the naming and help text appropriately.

As an aside, I think we should change the way we provision devices to omit the timestamp - I'm going to open a ticket for that with the cloud team.

samples/hello/scripts/generate-keys.py
sam-golioth: Good catch!

samples/hello/scripts/generate-keys.py
sam-golioth: The `project_name` and `project_id` are actually different. In the root certificate generation command, we already need the API key and goliothctl config path if we're doing in the upload step - in that case, we can use goliothctl to get the project-id by name. If we're not doing the upload in the script (i.e. we'll upload the root cert manually later), then we don't need the project name or project id. For the device generation command, there's no communication necessary with the Golioth servers (this is actually one of the main reason I split them into separate commands). In that case, you need to provide the project-id because we have no way of converting project-name to project-id.

I'll change the root generation command to accept project-id instead of project-name. It will be a little less convenient for the user, but I think the benefit from reducing confusion is greater. Thanks for the feedback!

samples/common/scripts/generate-keys.py
mniestroj: When I list devices with `goliothctl` I get the following:
```
id:"6033cc457016b281d671df53"  hardware_ids:"DE:AD:BE:EF"  name:"nrf52"
...
id:"63989e55811d6b602efec176"  hardware_ids:"20221213154629-de-ad-be-ef"  name:"DE:AD:BE:EF"
```
All my comments about were with the assumption that I was using device `hardware_ids:"DE:AD:BE:EF"  name:"nrf52"`... which seems that I was wrong?

https://www.notion.so/golioth/Device-Certificate-Creation-Guide-f6d859731c784128b7de33d152e11bdd#cde455fa44dd40b185b186ff1dd8f5a9 was used to develop cert-based auth. I am pretty sure that when asking verbally about what the "primary hardware id" means I got a response something along the line "it is the first hardware id".

So... I used `DE:AD:BE:EF` to generate certificates. Device connected. I probably used a `goliothctl` command that shows all the logs (since it was the only service verified after cert-based auth) from all devices. It looked like "my" (`name:"nrf52"`) device has connected. What happened instead is that new device was created and generated hardware-id: `hardware_ids:"20221213154629-de-ad-be-ef"  name:"DE:AD:BE:EF"`. I am surprised, just learned about it seconds ago.

This problem is not related now to this PR. We need to fix our documentation and do not cause any confusion. Question is: what was the intended behavior? Should the "name" be used or "hardware id"?

sam-golioth: Yeah, I think the CN should map to directly to the "primary hardware id", and also populate the name (just because the device needs a name when it gets created). And then users can go in and change the name in the console if they like.

That actually brings up another problem: that you can change both device name and hardware id on the console - which would break the association between a physical device and its entry in the console.

I agree, fixing this is outside of the scope of this PR.

mniestroj (commented):
Code changes look fine. However I would like to clarify the "device primary hardware id" vs "device name" thing. This is new to me, since I was 100% sure that so far a "hardware id" was used, not a "name".

sam-golioth:
> Code changes look fine. However I would like to clarify the "device primary hardware id" vs "device name" thing. This is new to me, since I was 100% sure that so far a "hardware id" was used, not a "name".

It surprised me too. Here's the relevant part of the backend code: https://github.com/golioth/golioth/blob/main/internal/services/deviceauth/certificates_auth.go#L113

PR #406, opened 2023-07-05
auth: move PSK handling to samples

sam-golioth:
- move system settings into samples

- add PSK support to hardcoded_credentials module

- remove PSK handling in system_client

- update samples to use the same auth for consistency


samples/settings/src/main.c
```diff
@@ -61,6 +62,14 @@ int main(void)
...
 	}
 
+	/* Note: In production, you would provision unique credentials onto each
+	 * device. For simplicity, we provide a utility to hardcode credentials as
+	 * kconfig options in the samples.
+	 */
+	if (IS_ENABLED(CONFIG_GOLIOTH_SAMPLE_HARDCODED_CREDENTIALS)) {
+		hardcoded_credentials_set();
+	}
+
```
mniestroj, line 72: How about calling `hardcoded_credentials_set()` automatically during boot by `SYS_INIT(..., APPLICATION, ...)`? That way we would not need to have that as part of every sample and would keep sample specific code minimal.

samples/common/Kconfig
```diff
@@ -80,7 +105,7 @@ config GOLIOTH_SAMPLE_WIFI
 
 config GOLIOTH_SAMPLE_WIFI_SETTINGS
 	bool "Load SSID and PSK from settigs subsystem"
-	default y if GOLIOTH_SYSTEM_SETTINGS
+	default n
```
mniestroj, line 108: `default n` is implicit, so no need to set that (this line can be removed). Also it is quite hard to which logical change this is related, since all changes are in single commit.

samples/hello/README.rst
```diff
@@ -19,29 +19,73 @@ Building and Running
...
+
+PSK based auth - Hardcoded
+--------------------------
 
 Configure the following Kconfig options based on your Golioth credentials:
 
-- GOLIOTH_SYSTEM_CLIENT_PSK_ID  - PSK ID of registered device
-- GOLIOTH_SYSTEM_CLIENT_PSK     - PSK of registered device
+- ``GOLIOTH_SAMPLE_HARDCODED_PSK_ID``  - PSK ID of registered device
+- ``GOLIOTH_SAMPLE_HARDCODED_PSK``     - PSK of registered device
```
mniestroj, line 34: Please do this kind of improvement in separate commit, so it is less overwhelming to review logical changes.

samples/hello/README.rst
```diff
@@ -184,6 +228,8 @@ This is the output from the serial console:
 Responses to Hello messages are printed above as a hexdump of "Hello mark". This
 means that communication with Golioth is working.
 
+.. _authentication methods: https://docs.golioth.io/firmware/zephyr-device-sdk/authentication/
```
mniestroj, line 231: No strong opinion here... but I'm a bit afraid of circular referencing (docs are written based this SDK, while here we start to reference those docs), since this page is strictly Golioth Zephyr SDK (this repo) related. It might quickly get outdated.

samples/hello/prj.conf
```diff
@@ -2,3 +2,9 @@ CONFIG_GOLIOTH_SAMPLES_COMMON=y
 
 # Application
 CONFIG_LOG_BACKEND_GOLIOTH=y
+
+CONFIG_GOLIOTH_SAMPLE_HARDCODED_PSK_ID="20230504202622-bar-lights-esp@playground-2f69a2"
+CONFIG_GOLIOTH_SAMPLE_HARDCODED_PSK="641edbbfe7c20d00fb3e4e87684b318f"
+
+CONFIG_GOLIOTH_SAMPLE_WIFI_SSID="WeWorkDevice"
+CONFIG_GOLIOTH_SAMPLE_WIFI_PSK="mousy sept glacier surtax"
```
mniestroj, line 10: Is this a leftover to be removed?

samples/settings/src/main.c
sam-golioth: I like it!

samples/hello/README.rst
sam-golioth: Good call out.

samples/hello/prj.conf
sam-golioth: Yes 🤦

samples/settings/prj.conf
```diff
@@ -3,17 +3,6 @@ CONFIG_GOLIOTH_SAMPLES_COMMON=y
...
-
-CONFIG_FLASH=y
-CONFIG_FLASH_MAP=y
-CONFIG_NVS=y
-
-CONFIG_SETTINGS=y
-CONFIG_SETTINGS_RUNTIME=y
-CONFIG_GOLIOTH_SAMPLE_SETTINGS_AUTOLOAD=y
-CONFIG_GOLIOTH_SAMPLE_SETTINGS_SHELL=y
-
```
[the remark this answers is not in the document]
szczys: > enable storing PSK-ID/PSK in settings subsystem, so that the DFU sample can be flashed on multiple devices as is and then do a DFU of the fleet

Yes, this is how precompiled binaries for the zephyr-training, thingy91demo, and all the reference designs currently work. We're using PSK auth but it's not hardcoded into the firmware.

net/golioth/Kconfig
```diff
@@ -285,17 +285,6 @@ config GOLIOTH_SYSTEM_CLIENT_RX_BUF_SIZE
...
-	bool "Load credentials from persistent settings"
-	default y
-	depends on GOLIOTH_AUTH_METHOD_PSK
-	depends on SETTINGS
-	help
-	  When selected, Golioth credentials will be loaded from settings
-	  subsystem.
-
-if GOLIOTH_SYSTEM_SETTINGS
-
```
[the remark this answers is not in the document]
szczys: Most notable in my mind is persistent storage for WiFI credentials. Currently we use the shell to set to those in the storage partition.

samples/settings/prj.conf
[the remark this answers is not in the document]
sam-golioth: Hardcoded PSKs are just the default. As part of this PR, there's a new overlay config file in `samples/common` that can enable runtime PSKs via the settings subsystem for _any_ of our samples, e.g.:
`west build -p -b esp32 samples/dfu -- -DOVERLAY_CONFIG="../common/runtime_psk.conf"`

So this should give us the best of both worlds, while maintaining consistency among our samples. All the samples use hardcoded PSKs by default, and all can be changed to use runtime PSKs and the settings shell with a single build argument. This is documented in the READMEs for each of the samples now, e.g.: https://github.com/golioth/golioth-zephyr-sdk/pull/406/files#diff-40c1e6da1ab07e4c146dae9c4f8ecf4c2dd4afdf703d11f21d10dc6b2e4898c3R46

net/golioth/Kconfig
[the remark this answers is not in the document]
mniestroj: Upstream Zephyr has now support for connection/network manager with some functionalities implemented like management of WiFi (or other medium type) interface. Such "generic" functionalities make most sense to be included as part of Zephyr, since those are not cloud-provider (Golioth) specific. We should make use of those as much as possible, so that:
a) we don't maintain too much utilities that are not directly related to Golioth as a platform
b) it makes less friction for users that want to integrate their existing (or under development) app into Golioth

In other words, it would be great to provide Golioth SDK users some utilities, even for downstream applications. But in those cases it makes more sense to upstream such feature and let it be improved by larger community. Development of such feature (if it is well-thought) should take the same time in both cases. So in most cases some example fits best in `samples/common/` IMO.

In case of PSK-ID/PSK TLS credentials, I think that the ideal situation would be to provide a "TLS credentials storage which utilizes Zephyr settings subsystem". The only thing that would be Golioth-specific is the TLS ID which is used to reference those credentials. This TLS ID is already configured in Kconfig. The rest of it make sense to be put in Zephyr upstream. We can create something temporarily here in SDK until a "proper" implementation is merged to Zephyr, so that we don't introduce any latency between development and being able to use it.

Settings shell is one example why above strategy with upstreaming would be beneficial. First a settings shell was created in Golioth SDK. Shortly after that a different implementation (but with similar functionality in mind) was merged to Zephyr. Then there was a conflict between Kconfig symbols. It was solved (by prepending "our" service with `GOLIOTH_` in Kconfig option name). Right now there is still a conflict in shell module, so either one is visible or the other, since both commands are named `settings`.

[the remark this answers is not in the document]
sam-golioth: I like this a lot! I agree, upstreaming is the way to go. 

I filed https://github.com/golioth/firmware-issue-tracker/issues/183 and https://github.com/golioth/firmware-issue-tracker/issues/184 for some of the work that you mentioned, feel free to edit/add/delete.

> We can create something temporarily here in SDK until a "proper" implementation is merged to Zephyr

Where should that live? Still `samples/common` or do we make another folder for things that are in the process of being upstreamed?

mniestroj (approved):
LGTM, in terms of code changes.

Would be good to have @golioth/developer-relations approval (doesn't have to be here).

sam-golioth:
> LGTM, in terms of code changes.
> 
> Would be good to have @golioth/developer-relations approval (doesn't have to be here).

Confirmed with @ChrisGammell

PR #412, opened 2023-07-21
bump Zephyr to tip of main

mniestroj:
Update to the newest version of Zephyr. This requires handling of DORMANT network interface states,
which is handled in this PR as well.


samples/common/net_connect.c
```diff
@@ -73,8 +73,11 @@ void net_connect(void)
 {
 	struct net_if *iface = net_if_get_default();
 
-	LOG_INF("Waiting for interface to be up");
-	wait_for_iface_up(iface);
+	if (IS_ENABLED(CONFIG_GOLIOTH_SAMPLE_WIFI) &&
```
sam-golioth, line 76: We can combine this with the block below at line 82 to make it a little cleaner.

sam-golioth (approved):

samples/common/net_connect.c
mniestroj: Actually implemented check was broken. It was supposed to skip waiting for IFACE UP event for "WiFi && Dormant" interfaces, but still wait for IFACE UP for all others. Fixed that now. After the fix your comment about combining with block at 82 line does not apply anymore. But thanks for commenting about it, it allowed me to notice this bug. Tested with `mimxrt1060_evkb` and confirmed it works after the fix.

mniestroj:
There are problems with ESP32 platform and DFU, hence marked this PR as `DNM`. Working on that now.

mniestroj:
Went with another bump of Zephyr. This pulls in `esp32` board name change to `esp32_devkitc_wroom`. All the bumps are in separate commits, as there are separate issues targeted in each. This will also allow to easily bisect between those bumps in case we find regressions because of some of them.

mniestroj:
> Should we consider adding support for other ESP32 boards, now that they're split up? Specifically thinking about the ESP32-DevKitC-WROVER. It can be a separate change.

I would like to keep the amount of supported platforms minimal. We have reduced those in the past, just because it was consuming to much time to make sure every platform worked. Second thing is that I don't even have this HW right now. We can consider more HWs if we make improvements in the HIL in CI and make all the tests autonomous. Right now we are not there yet.

PR #420, opened 2023-08-04
samples: Use EXTRA_CONF_FILE instead of OVERLAY_CONFIG

sam-golioth:
OVERLAY_CONFIG was deprecated in favor of EXTRA_CONF_FILE in 6a98c198d8f3

### DIFF
--- a/samples/dfu/README.rst
+++ b/samples/dfu/README.rst
@@ -54,7 +54,7 @@ file:
 
 .. code-block:: console
 
-   $ west build -- -DOVERLAY_CONFIG=${ZEPHYR_GOLIOTH_MODULE_DIR}/samples/common/runtime_psk.conf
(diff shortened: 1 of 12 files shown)


sam-golioth:
@mniestroj One more for v0.7.0. Feel free to merge this if it looks good to you and you're working on the release before I get online Monday.

mniestroj (commented):
Changes look good, but what is `6a98c198d8f3` referenced in commit body? I don't think it is a valid commit sha1, if that is what it was supposed to be.

sam-golioth:
> Changes look good, but what is `6a98c198d8f3` referenced in commit body? I don't think it is a valid commit sha1, if that is what it was supposed to be.

Whoops. It was supposed to be Zephyr upstream, but I think I must have taken the commit from the PR branch instead of `main`. It should be `3a345682ba81`. I'll update the commit message with the right SHA1 and to indicate that it's from Zephyr, not this repo.

mniestroj (approved):

PR #423, opened 2023-08-09
fw: inform user when no firmware release found

szczys:
When no firmware release is available for a device, Golioth will return a zero-length CoAP map (0xa0). When this value is received, show a warning message that "No fw release available", instead of the default "Failed to decode desired map"


net/golioth/fw.c
```diff
@@ -123,6 +123,10 @@ int golioth_fw_desired_parse(const uint8_t *payload, uint16_t payload_len,
 
 	err = zcbor_map_decode(zsd, map_entries, ARRAY_SIZE(map_entries));
 	if (err) {
+		if ((payload_len == 1) && (payload[0] == 0xa0)) {
```
szczys, line 126: Is there a macro for CoAP zero-length map? It would be nice to use that instead of the `0xa0` literal but I wasn't able to find it.

szczys (commented):
I know of at least one user that was blocked by this issue, receiving 0xa0 but not understanding why. I think it would be useful to have a message when we know the `desired` value indicates no releases.

net/golioth/fw.c
```diff
@@ -123,6 +123,10 @@ int golioth_fw_desired_parse(const uint8_t *payload, uint16_t payload_len,
 
 	err = zcbor_map_decode(zsd, map_entries, ARRAY_SIZE(map_entries));
 	if (err) {
+		if (payload_len == 1 && payload[0] == 0xa0) {
```
szczys, line 126: Is there a macro for CoAP zero-length map? It would be nice to use that instead of the `0xa0` literal but I wasn't able to find it.

mniestroj:
Good catch! It is definitely a regression. @szczys What do you think about #424 ? It would basically bring the behavior we have previously when QCBOR was used.

szczys:
Yes, #424 looks like a better solution!

PR #426, opened 2023-08-11
zcbor: backport zcbor_any_skip() for zcbor < 0.7.0

mniestroj:
Backport fixed `zcbor_any_skip()` from zcbor 0.7.0.

This fixes 'samples/rpc/' when used with NCS, which pulls zcbor in version
0.6.0.

Fixes: https://github.com/golioth/firmware-issue-tracker/issues/197


mniestroj:
All checkpatch errors are related to code copy-pasted from zcbor library. Since this is a temporary solution (until there is new release of NCS), IMO we should just ignore those for the sake of this PR.

sam-golioth:
> All checkpatch errors are related to code copy-pasted from zcbor library. Since this is a temporary solution (until there is new release of NCS), IMO we should just ignore those for the sake of this PR.

Especially because it's contained to one file.

sam-golioth (approved):

PR #429, opened 2023-08-30
Samples: Certificate Provisioning: add nrf52840

szczys:
* [samples: cert-provisioning: change log module name](https://github.com/golioth/golioth-zephyr-sdk/commit/a5e81e59cdf88eff6ed2f0d919f44e6194459daf)
* [samples: cert-provsioning: add nRF52840](https://github.com/golioth/golioth-zephyr-sdk/commit/c0331f4e000a3468705bf5d3e7ebe755af034e6a)


samples/certificate_provisioning/boards/nrf52840dk_nrf52840.conf
```diff
@@ -0,0 +1,18 @@
+# WiFi
+CONFIG_WIFI=y
+CONFIG_WIFI_ESP_AT=y
+CONFIG_WIFI_ESP_AT_MDM_RX_BUF_COUNT=40
+
+CONFIG_NET_L2_WIFI_SHELL=y
+CONFIG_GOLIOTH_SAMPLE_WIFI=y
+
+# General config
+CONFIG_HEAP_MEM_POOL_SIZE=4096
```
szczys, line 10: I chose this value arbitrarily from the nRF9160dk conf file. This symbol is necessary for the file system:

```
warning: FILE_SYSTEM_SHELL (defined at subsys/fs/Kconfig:46) was assigned the value 'y' but got the
value 'n'. Check these unsatisfied dependencies: (HEAP_MEM_POOL_SIZE > 0) (=n). See

```

But is there a way to know what size is necessary?

samples/certificate_provisioning/src/main.c
```diff
@@ -5,7 +5,7 @@
  */
 
 #include <zephyr/logging/log.h>
-LOG_MODULE_REGISTER(golioth_hello, LOG_LEVEL_DBG);
+LOG_MODULE_REGISTER(cert_provisioning, LOG_LEVEL_DBG);
```
sam-golioth, line 8: 🙏

samples/certificate_provisioning/README.rst
```diff
@@ -48,6 +48,60 @@ sample application (i.e. ``samples/certificate_provisioning``) and type:
...
+
+.. code-block:: console
+
+   esptool.py write_flash --verify 0x0 PATH_TO_ESP_AT/factory/factory_WROOM-32.bin
+
+Connect nRF52840 DK and ESP32-DevKitC V4 (or other ESP32-WROOM-32 based board)
+using wires:
+
++-----------+--------------+
+|nRF52840 DK|ESP32-WROOM-32|
```
sam-golioth, line 68: Should we add WROVER pins while we're at it? Or maybe a separate PR adding it for all the samples makes more sense.

samples/certificate_provisioning/boards/nrf52840dk_nrf52840.conf
```diff
@@ -0,0 +1,18 @@
...
+
+# MCU Manager
+CONFIG_MCUBOOT_BOOTUTIL_LIB=y
+CONFIG_MCUMGR_MGMT_NOTIFICATION_HOOKS=y
+CONFIG_MCUMGR_GRP_FS_FILE_ACCESS_HOOK=y
```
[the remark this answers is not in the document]
szczys: These symbols were added to remove this compiler warning that appears when the device is flashed (I previously tried to replicate it during the build step but it's only present during the flashing step for some baffling reason):

```
CMake Warning at /home/mike/golioth-zephyr-workspace/zephyr/subsys/mgmt/mcumgr/grp/fs_mgmt/CMakeLists.txt:26 (message):
  Note: MCUmgr file system management is enabled but file access hooks are
  disabled, this is an insecure configuration and not recommended for
  production use, as all files on the filesystem can be manipulated by a
  remote device.  See
  https://docs.zephyrproject.org/latest/services/device_mgmt/mcumgr_callbacks.html
  for details on enabling and using MCUmgr hooks.
```

I haven't seen this message when using the nRF9160dk so I thought it best to put it in the board conf.

Would it help to use a different comment? Like:

```
# Use secure configuration file access options for nRF52840
```

samples/certificate_provisioning/README.rst
szczys: I added the WROVER entry to this one and created #430 for the rest. Please review the new RST syntax I used on the table.

samples/certificate_provisioning/boards/nrf52840dk_nrf52840.conf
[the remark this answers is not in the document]
sam-golioth: I'm not sure why the warning is only showing up when flashing the nrf52840dk, but it applies to all boards. For a sample, there's not really any problem with this warning, other than being annoying to users. In fact, even if it weren't a sample there wouldn't be an issue, because the only files in the filesystem are ones that we want to expose over `mcumgr`. Blocking access to other files doesn't actually make the device more secure (for this application).

We can fix it or we can leave it. I'm of the opinion that the warning is fine, and we should just ignore it, but it's not a strong opinion. If we decide to fix it, we should not only enable `CONFIG_MCUMGR_MGMT_NOTIFICATION_HOOKS` and `CONFIG_MCUMGR_GRP_FS_FILE_ACCESS_HOOK` in `prj.conf`, but actually implement the fs access hook to only allow access to the files in question, for completeness. Enabling the hook but not implementing is equivalent to not enabling it, save for silencing the warning, and doing so could set a bad example for users. The reason I think we should leave the warning is because the point of the sample is _not_ to show how to use `mcumgr` in general, and adding this extra code could unnecessarily complicate the sample. And if we decide to fix it, it should probably be in a separate ticket/PR, since this affects all boards.

Separate question: What do we need `CONFIG_MCUBOOT_BOOTUTIL_LIB` for?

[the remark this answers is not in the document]
szczys: Okay, I agree we should leave these symbols out and be okay with the warning.

> Separate question: What do we need CONFIG_MCUBOOT_BOOTUTIL_LIB for?

I think I made a mistake when looking up dependencies. I think I was searching for hooks and looked at `BOOT_IMAGE_ACCESS_HOOKS` by mistake which depends on `MCUBOOT_BOOTUIL_LIB`. Sorry about that.

samples/certificate_provisioning/boards/nrf52840dk_nrf52840.conf
sam-golioth: I'm thinking maybe we should move this symbol to `prj.conf` too, since all the boards will need the file system shell.

sam-golioth (approved):

PR #433, opened 2023-10-05
bump Zephyr to v3.5.0-rc1 and NCS to tip of main

mniestroj:
v3.5.0-rc1 is the newest release candidate. Update now, so there are more
chances to discover potential bugs.

Bump NCS to tip of main, so that twister implementation is almost on par
with vanilla Zephyr 3.5.0-rc1. Drop old cherry-picked patches and apply 3
newer. This allows to keep the same pytest test scripts for both vanilla
Zephyr and NCS.


mniestroj:
`checkpatch` warnings are false-positives, once again. This is related to `.patch` files being removed/added, which seems to  confuse `checkpatch.pl` script.

PR #434, opened 2023-10-06
samples: implement WiFi state machine to handle reconnects

mniestroj:
Implement WiFi connectivity state machine running in a dedicated work
queue (which runs in a dedicated thread). Network management events from
Zephyr are handled there, as well as timeouts and retries when attempting
to connect. As a result, WiFi connectivity is bringed back whenever there
was a temporary problem with WiFi AP connection.

In case WiFi radio is not ready (L1 is down, which results in LOWER_DOWN
interface management event/state) wait until it becomes ready (which is
required by ESP32 WiFi driver, since it does not handle connect requests
otherwise).

Initialize and launch WiFi work queue thread automatically by using
SYS_INIT(), so that `net_connect()` function does not need to handle WiFi
explicitly.

Since `net_connect()` handles mostly DHCPv4 setup and waiting to obtain
IPv4 address, there is no need to check whether interace is UP at the
beginning. DHCPv4 in Zephyr has its own state machine, with network events
handling, all it requires is just calling net_dhcpv4_start() even for
interfaces that are not operational at the time of function call.

Replace waiting for "DHCPv4 bound" event with waiting for "IPv4 address
add" event. The latter allows to wait for WiFi to be connected to AP with
assigned IP address, while still handling (implicitly) "DHCPv4 bound"
event, which is required for ethernet cable connected devices.


samples/common/wifi.c
```diff
@@ -128,54 +254,287 @@ void wifi_connect(struct net_if *iface)
...
+{
+	struct wifi_manager_data *wifi_mgmt =
+		CONTAINER_OF(work, struct wifi_manager_data, event_work);
+	enum wifi_event event = wifi_mgmt->event;
+	enum wifi_state old_state = wifi_mgmt->state;
+	int map_value = wifi_state_change_map[event][old_state];
+	enum wifi_state new_state = map_value - MAP_OFFSET;
+	int sleep_msec = 0;
+
+	if (map_value - MAP_OFFSET >= WIFI_STATE_WAIT) {
```
sam-golioth, line 365: That's a neat trick

samples/common/wifi.c
```diff
@@ -1,44 +1,143 @@
...
+
+#define IDL		(WIFI_STATE_IDLE + MAP_OFFSET)
+#define LUP		(WIFI_STATE_WAITING_FOR_LOWER_UP + MAP_OFFSET)
+#define CON		(WIFI_STATE_CONNECTING + MAP_OFFSET)
+#define DIS		(WIFI_STATE_DISCONNECTING + MAP_OFFSET)
+#define RDY		(WIFI_STATE_READY + MAP_OFFSET)
+#define INV		(WIFI_STATE_INVALID + MAP_OFFSET)
+#define WAIT(ms)	(WIFI_STATE_WAIT + MAP_OFFSET + (ms))
+
+static const int wifi_state_change_map[][WIFI_STATE_LAST + 1] = {
```
sam-golioth, line 131: Why not use the [State Machine Framework](https://docs.zephyrproject.org/latest/services/smf/index.html)?

mniestroj: I looked briefly at that framework months ago and... I was just expecting something more I guess. With a hierarchical state machine this might be a good thing to use, but for flat state machine I got the impression that it didn't provide that much tools to speed up development or reuse.

I guess that the real answer could be provided only by trying to reimplement what we have in this PR with State Machine Framework and see how it looks in practice. This approach should be faster and give more precise answer than theoretical considerations.

sam-golioth (approved):

PR #435, opened 2023-10-06
Implement GOLIOTH_SAMPLE_FW_VERSION

szczys:
This implements the `GOLIOTH_SAMPLE_FW_VERSION` symbol for setting the firmware version in the prj.conf file. This matches what we are doing in the Firmware SDK.

## Considerations

* This symbol makes the versioning process the same for both Zephyr and NCS
* Firmware version can still be overriden at build time using `-DCONFIG_GOLIOTH_SAMPLE_FW_VERSION=\"1.2.4\"`
* I have a Docs PR coming to match these changes

I tested the DFU sample with these changes on the following boards:

* nRF9160dk
* nRF52840 + ESP32
* esp32_devkitc_wroom

##  Notable

I don't think this will affect the twister testing. I looked at the approach in the pytest files and it seems that the `CONFIG_MCUBOOT_IMAGE_VERSION` is being patched directly once the `.config` file is already generated. However, I have not run the Twister test on these changes.


samples/dfu/README.rst
```diff
@@ -108,12 +108,9 @@ Build Zephyr sample application for nRF9160 DK:
    $ west build -b nrf9160dk_nrf9160_ns samples/dfu
    $ west flash
 
-Now rebuild the application with the new version number 1.2.3 to distinguish it
-from the old firmware:
-
-.. code-block:: console
-
-   $ west build -b nrf9160dk_nrf9160_ns samples/dfu -- -DCONFIG_MCUBOOT_IMAGE_VERSION=\"1.2.3\"
```
sam-golioth, line 116: We should still be able to override the `CONFIG_GOLIOTH_SAMPLE_FW_VERSION` symbol from the `west build` command right? I don't think we need to document it as part of the sample, but that could be useful for clients (for example in CI systems where the version is automatically incremented), so just checking that the functionality remains.

sam-golioth (approved):

samples/common/Kconfig
```diff
@@ -60,6 +60,25 @@ config GOLIOTH_SAMPLE_DHCP_BIND
 	help
 	  Obtain an IPv4 address from the DHCP server.
 
+config GOLIOTH_SAMPLE_FW_VERSION
```
mniestroj, line 63: After NCS bump to version "post v2.4.99-dev2" there is `CONFIG_MCUBOOT_IMGTOOL_SIGN_VERSION` used in NCS (which is the same symbol as in vanilla Zephyr). This means that main (I guess) reason why `GOLIOTH_SAMPLE_FW_VERSION` was introduced is no longer valid.

Additionally there is a way to define application version with Zephyr infrastructure, by placing `VERSION` file in application directory. See https://github.com/zephyrproject-rtos/example-application/commit/2c85d9224fca21fe6e370895c089a6642a9505ea for example usage. Not sure this is viable option for our SDK, since we just provide information on how to set version from commandline in order to quickly test that feature, not permanently because we release application.

Since there are already 2 ways (see above) how (application) version can be set, I don't see good enough reason to introduce another way.

mniestroj (changes requested):

PR #438, opened 2023-10-17
bump Zephyr to 3.5.0-rc3 and NCS to tip of main (2.5.0-rc1+)

mniestroj:
Bump Zephyr to latest release candidate to test it before final 3.5.0
release.

Bump NCS to tip of main (slightly newer than 2.5.0-rc1 tag), to find
potential nRF91* platform regressions.

Replace `rand32.h` with `random.h` to suppress header deprecation warning. Add
a workaround for older Zephyr versions (like the one used by NCS fork) by
providing `random.h` just for internal Golioth SDK library visibility.

Tested on:
- [x] `qemu_x86`
- [x] `nrf52840dk_nrf52840`
- [x] `mimxrt1060_evkb`
- [x] `esp32_devkitc_wroom`
- [x] `nrf9160dk_nrf9160_ns`

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -26,7 +26,7 @@ jobs:
           - qemu_x86
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           path: modules/lib/golioth
 
@@ -67,7 +67,7 @@ jobs:
           -T modules/lib/golioth
 
       - name: Upload artifacts
-        uses: actions/upload-artifact@v2
+        uses: actions/upload-artifact@v3
         if: ${{ always() }}
         with:
           name: twister-artifacts
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -17,6 +17,14 @@ jobs:
     env:
       ZEPHYR_SDK_INSTALL_DIR: /opt/toolchains/zephyr-sdk-0.16.3
 
+    strategy:
+      fail-fast: false
+      matrix:
+        board:
+          - esp32_devkitc_wroom
+          - nrf52840dk_nrf52840
+          - qemu_x86
+
     steps:
       - uses: actions/checkout@v3
         with:
@@ -45,9 +53,7 @@ jobs:
           zephyr/scripts/twister
           --no-clean
           -e goliothd
-          -p esp32_devkitc_wroom
-          -p nrf52840dk_nrf52840
-          -p qemu_x86
+          -p ${{ matrix.board }}
           -o reports/non_goliothd
           -T modules/lib/golioth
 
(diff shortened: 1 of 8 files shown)


sam-golioth:
Looks like the [build test failed](https://github.com/golioth/golioth-zephyr-sdk/actions/runs/6547204552?pr=438), because the runner ran out of disk space. I hit the same error in the CP SDK, and you should be able to fix it by parallelizing the build similar to https://github.com/golioth/golioth-firmware-sdk/commit/5eb86b75ba9e9561d6fe013ef69b793eccf5301d

sam-golioth (changes requested):
Looks like the [build test failed](https://github.com/golioth/golioth-zephyr-sdk/actions/runs/6547204552?pr=438), because the runner ran out of disk space. I hit the same error in the CP SDK, and you should be able to fix it by parallelizing the build similar to https://github.com/golioth/golioth-firmware-sdk/commit/5eb86b75ba9e9561d6fe013ef69b793eccf5301d

mniestroj:
> Looks like the [build test failed](https://github.com/golioth/golioth-zephyr-sdk/actions/runs/6547204552?pr=438), because the runner ran out of disk space. I hit the same error in the CP SDK, and you should be able to fix it by parallelizing the build similar to [golioth/golioth-firmware-sdk@5eb86b7](https://github.com/golioth/golioth-firmware-sdk/commit/5eb86b75ba9e9561d6fe013ef69b793eccf5301d)

Proposed solution (https://github.com/golioth/golioth-zephyr-sdk/pull/439) is slightly different, much smaller in boilerplate code size, but seems like a perfect fit for what we have now. This will give some input for comparison with the solution you proposed and implemented for CP SDK. Both are fine, maybe they can be combined together. My implementation was inspired by Zephyr CI matrix (but they use arbitrary number of matrix instances, which is then passed to twister, which decides by its own how to divide all test cases to be executed).

mniestroj:
@sam-golioth Do you have some idea about why `build` job is "Expected - Waiting for status to be reported"? It looks like a job that is the parent for 3 matrix jobs. Those 3 jobs are successful, but still, `build` is somehow pending. In case that is something new, I'll just investigate what the issue might be (I guess something in workflow specification which is not exactly what GitHub expects, or I just need close/reopen to workaround some corner case in GitHub CI).

sam-golioth:
> @sam-golioth Do you have some idea about why `build` job is "Expected - Waiting for status to be reported"?

I had marked the `build` job as a required check after it failed on this PR earlier. The reason is because when a runner runs out of disk space, the job is marked as "skipped" not "failed" and the PR UI treats skipped jobs like successful jobs (it still said "All checks have passed"), so it wasn't obvious that there was an issue. Marking it as required makes sure we don't miss when a job is "skipped". It looks like there isn't a way to mark the parent job as required, so I've just removed the requirement. We can mark the child jobs as required later, after this gets merged.

Also, the job being marked as "skipped" and not "failed" might be due to something in our workflow rather than an unsolvable GitHub actions issue, but I didn't want to put in the time to investigate that right now. Just something to keep in mind for the future.

sam-golioth (approved):

PR #440, opened 2023-10-19
github-ci: use matrix to parallelize jobs

mniestroj:
Use matrix of boards to parallelize jobs across multiple runners. This will also reduce amount of disk space needed at the end.

Clean push of https://github.com/golioth/golioth-zephyr-sdk/pull/439


mniestroj:
Merged as part of #438

PR #443, opened 2023-10-23
scripts: install python golioth package from github

sam-golioth:
The golioth python package at scripts/python/golioth has been moved into [its own git repository](https://github.com/golioth/python-golioth-tools). This commit deletes the local copy and changes gitlab ci to install the package from git instead of from the local copy.

Resolves golioth/firmware-issue-tracker#271


.gitlab-ci.yml
```diff
@@ -140,7 +140,7 @@ twister-qemu-goliothd:
       EOF
     # Install samples/tests requirements
     - pip3 install -r modules/lib/golioth/scripts/requirements-tests.txt
-    - pip3 install modules/lib/golioth/scripts/python/golioth
+    - pip3 install git+https://github.com/golioth/python-golioth-cli
```
mniestroj, line 143: So far scripts were part of the repo. As a result git repo itself made the contract that specific version of scripts matched specific pytest/twister scripts implementations.

With external repo for Python scripts we need to make a contract on the API level in order to make sure to (or at least try to) test old as well as new versions of firmware (and corresponding pytest/twister scripts). In other words, we need some versioning of the Python scripts and not rely on the tip of `main` branch here.

Additionally `python-golioth-cli` name would be fine for CLI tools only, but this package will actually provide Python APIs for HTTP REST API exposed by Golioth cloud. I would rather remove `-cli` from the name and use something like `python-golioth-api`.

mniestroj (changes requested):

.gitlab-ci.yml
sam-golioth: I changed the name of the repo to `python-golioth-tools` (but happy to take other suggestions as well). I also tagged the tip of main as `v0.1.0` and changed the pip install command to reference that tag.

mniestroj (approved):

PR #444, opened 2023-10-30
samples: enable wifi settings in runtime settings conf overlay

sam-golioth:
Enable wifi settings shell when using the runtime settings config
overlay file. Also rename the config overlay from `runtime_psk.conf`
to `runtime_settings.conf`.

Closes golioth/firmware-issue-tracker#282


samples/common/Kconfig
```diff
@@ -123,6 +123,7 @@ config GOLIOTH_SAMPLE_WIFI_SETTINGS
 	bool "Load SSID and PSK from settigs subsystem"
 	depends on GOLIOTH_SAMPLE_WIFI
 	depends on SETTINGS
+	default y if GOLIOTH_SAMPLE_PSK_SETTINGS
```
mniestroj, line 126: I am not sure why PSK settings should enable (by default) WiFi settings. Those options sound like two independent features.

sam-golioth: It's more about user expectation, but I agree they aren't related. I'll enable this option in `samples/common/runtime_psk.conf` instead.

mniestroj (approved):

PR #445, opened 2023-10-30
ci: use minimal docker image for build test

sam-golioth:
Use a Golioth created docker image for build test instead of Zephyr project image. The Golioth image only contains the tools that we need to build our supported boards, so it is smaller and faster than the Zephyr image.

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -11,8 +11,7 @@ jobs:
   build:
     runs-on: ubuntu-latest
 
-    container:
-      image: zephyrprojectrtos/ci:v0.26.5
+    container: golioth/golioth-zephyr-base:0.16.3-SDK-v0
 
     env:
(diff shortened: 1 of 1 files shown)


.github/workflows/build_test.yml
```diff
@@ -44,6 +43,12 @@ jobs:
           git config --global user.name "Git User"
           west patch
 
+          pip3 install -r zephyr/scripts/requirements-base.txt
+          pip3 install -r zephyr/scripts/requirements-build-test.txt
+          pip3 install -r zephyr/scripts/requirements-run-test.txt
+          # Needed for TF-M
+          pip3 install cryptography pyasn1 pyyaml cbor>=1.0.0 imgtool>=1.9.0 jinja2 click
```
mniestroj, line 50: Can we make this list in alphabetical order? I'm okay keeping it as is, if there is a reason for it (like being copy-pasted from other place).

PR #446, opened 2023-10-30
dtls: enable DTLS 1.2 Connection IDs

sam-golioth:
Enable Connection ID support, which enables deviecs to maintain their connection with the server for longer periods and across IP address changes, resulting in lower bandwidth and reduced power usage.

### DIFF
--- a/net/golioth/Kconfig
+++ b/net/golioth/Kconfig
@@ -27,6 +27,15 @@ module = GOLIOTH
 module-str = golioth
 source "subsys/logging/Kconfig.template.log_config"
 
+config GOLIOTH_USE_CONNECTION_ID
+	bool "Use DTLS 1.2 Connection IDs"
+	select MBEDTLS_SSL_DTLS_CONNECTION_ID
+	help
+	  Use DTLS 1.2 Connection IDs. Connection IDs replace IP
+	  addresses as the session identifier, and can be used to
+	  reduce the number of handshakes a device has to make in
+	  certain scenarios.
+
 choice GOLIOTH_AUTH_METHOD
 	prompt "Authentication method support"
 
--- a/net/golioth/golioth.c
+++ b/net/golioth/golioth.c
@@ -90,6 +90,18 @@ static int golioth_setsockopt_dtls(struct golioth_client *client, int sock,
 		}
 	}
 
+	/* If Connection IDs are enabled, set socket option to send CIDs, but not require that the
+	 * server sends one in return.
+	 */
+#ifdef CONFIG_GOLIOTH_USE_CONNECTION_ID
+	int enabled = 1;
+
+	ret = zsock_setsockopt(sock, SOL_TLS, TLS_DTLS_CID, &enabled, sizeof(enabled));
+	if (ret < 0) {
+		return -errno;
+	}
+#endif /* CONFIG_GOLIOTH_USE_CONNECTION_ID */
+
 	return 0;
 }
 
--- a/samples/hello/sample.yaml
+++ b/samples/hello/sample.yaml
@@ -56,6 +56,19 @@ tests:
       - CONFIG_MBEDTLS_CIPHER_CCM_ENABLED=y
       - CONFIG_MBEDTLS_CIPHER_MODE_CBC_ENABLED=n
       - CONFIG_GOLIOTH_CIPHERSUITES="TLS_PSK_WITH_AES_128_CCM_8"
+  sample.golioth.hello.psk.fast.cid:
+    platform_allow: >
+      esp32_devkitc_wroom
+      mimxrt1060_evkb
+      nrf52840dk_nrf52840
+      qemu_x86
+    extra_configs:
+      - CONFIG_GOLIOTH_USE_CONNECTION_ID=y
+  sample.golioth.hello.psk.long_start.cid:
+    platform_allow: nrf9160dk_nrf9160_ns
+    timeout: 120
+    extra_configs:
+      - CONFIG_GOLIOTH_USE_CONNECTION_ID=y
   sample.golioth.hello.psk.long_start:
     platform_allow: nrf9160dk_nrf9160_ns
     timeout: 120


hasheddan (commented):
@sam-golioth will also need to set the socket option here:

https://github.com/golioth/golioth-zephyr-sdk/blob/9455d286ed4c4159f0f1dcbcf5db5e5ef4534dd7/net/golioth/golioth.c#L53

See similar change in CP SDK:

https://github.com/golioth/golioth-firmware-sdk/pull/163/files#diff-410a63a31cf0ae75a5a9bb14ba25e7781fbb87aa7be322035d516647a2b6a11aR218

sam-golioth:
> @sam-golioth will also need to set the socket option here:
> 
> https://github.com/golioth/golioth-zephyr-sdk/blob/9455d286ed4c4159f0f1dcbcf5db5e5ef4534dd7/net/golioth/golioth.c#L53
> 
> See similar change in CP SDK:
> 
> https://github.com/golioth/golioth-firmware-sdk/pull/163/files#diff-410a63a31cf0ae75a5a9bb14ba25e7781fbb87aa7be322035d516647a2b6a11aR218

Fixed! Thanks for the catch!

net/golioth/Kconfig
```diff
@@ -9,6 +9,7 @@ config GOLIOTH
 	depends on NETWORKING
 	select COAP
 	select MBEDTLS
+	select MBEDTLS_SSL_DTLS_CONNECTION_ID
```
mniestroj, line 12: > enables deviecs to maintain their connection with the server for longer periods ... 

I'd like to have some evidence that backs that up, or at least theoretical use-case that enables "longer periods" of connection.

> ... and across IP address changes

What is the device use-case/scenario in which IP address changes? Isn't it like <1% of the possible cases? Personally I haven't notice IP address change at all and I can't even make up a scenario when that would happen. In case of cellular connectivity I guess that BTS roaming should solve that problem with maintaining single IP address (otherwise we would all be having connectivity problems when using mobile phones, where majority of connections/streams are TCP-based).

> resulting in lower bandwidth and reduced power usage.

I'd like to see some evidence (at least theoretical) that backs that up. I've done some tests with `qemu_x86` and `lightdb/get` sample, which shows that enabling CID support as done in this commit is **increasing** network usage (and as a result increasing power consumption). Below data shows almost 20% more data usage when sending empty CoAP messages, but there is almost 15% increase when issuing simple LightDB GET requests on `/counter` LightDB value.

<details>
<summary> No CID

`7	0.014224278	192.0.2.1	192.0.2.2	CoAP	83	NON, MID:32497, Empty Message`

</summary>

```
Frame 7: 83 bytes on wire (664 bits), 83 bytes captured (664 bits) on interface tap0, id 0
Ethernet II, Src: ICANNIAN_00:53:3d (00:00:5e:00:53:3d), Dst: 4e:a9:94:1c:22:15 (4e:a9:94:1c:22:15)
Internet Protocol Version 4, Src: 192.0.2.1, Dst: 192.0.2.2
User Datagram Protocol, Src Port: 47964, Dst Port: 5684
    Source Port: 47964
    Destination Port: 5684
    Length: 49
    Checksum: 0x8b3a [correct]
    [Checksum Status: Good]
    [Stream index: 0]
    [Timestamps]
    UDP payload (41 bytes)
Datagram Transport Layer Security
    DTLSv1.2 Record Layer: Application Data Protocol: Constrained Application Protocol
        Content Type: Application Data (23)
        Version: DTLS 1.2 (0xfefd)
        Epoch: 1
        Sequence Number: 1
        Length: 28
        Encrypted Application Data: 000100000000000120b9fceb9a7952e41b27fe19373001761d0143fe
        [Application Data Protocol: Constrained Application Protocol]
Constrained Application Protocol, Non-Confirmable, Empty Message, MID:32497
    01.. .... = Version: 1
    ..01 .... = Type: Non-Confirmable (1)
    .... 0000 = Token Length: 0
    Code: Empty Message (0)
    Message ID: 32497
```

</details>

<details>
<summary> With CID

`7	0.014537102	192.0.2.1	192.0.2.2	CoAP	103	NON, MID:44687, Empty Message`

</summary>

```
Frame 7: 103 bytes on wire (824 bits), 103 bytes captured (824 bits) on interface tap0, id 0
Ethernet II, Src: ICANNIAN_00:53:05 (00:00:5e:00:53:05), Dst: 4e:a9:94:1c:22:15 (4e:a9:94:1c:22:15)
Internet Protocol Version 4, Src: 192.0.2.1, Dst: 192.0.2.2
User Datagram Protocol, Src Port: 50091, Dst Port: 5684
    Source Port: 50091
    Destination Port: 5684
    Length: 69
    Checksum: 0x7ce1 [correct]
    [Checksum Status: Good]
    [Stream index: 0]
    [Timestamps]
    UDP payload (61 bytes)
Datagram Transport Layer Security
    DTLSv1.2 Record Layer: Application Data Protocol: Constrained Application Protocol
        Special Type: Connection ID (25)
        Version: DTLS 1.2 (0xfefd)
        Epoch: 1
        Sequence Number: 1
        Connection ID: 57fb69307e4d46b0
        Length: 40
        [Content Type: Application Data (23)]
        Encrypted Application Data: 000100000000000155fded71cbdeb98716d74510d6218ffcd2942f0ab7608e3caf2b9e18…
        [Application Data Protocol: Constrained Application Protocol]
Constrained Application Protocol, Non-Confirmable, Empty Message, MID:44687
    01.. .... = Version: 1
    ..01 .... = Type: Non-Confirmable (1)
    .... 0000 = Token Length: 0
    Code: Empty Message (0)
    Message ID: 44687
```

</details>

net/golioth/golioth.c
```diff
@@ -90,6 +90,18 @@ static int golioth_setsockopt_dtls(struct golioth_client *client, int sock,
 		}
 	}
 
+	/* If Connection IDs are enabled, set socket option to send CIDs, but not require that the
+	 * server sends one in return.
+	 */
+#ifdef CONFIG_MBEDTLS_SSL_DTLS_CONNECTION_ID
+	int ENABLED = 1;
```
mniestroj, line 97: I don't like `ENABLED` in uppercase, since I haven't seen such pattern yet. I'd rather use lowercase, just as any other variables and leave uppercase symbols for C preprocessor macros.

mniestroj (changes requested):

net/golioth/Kconfig
hasheddan: > I'd like to have some evidence that backs that up, or at least theoretical use-case that enables "longer periods" of connection.

@mniestroj here is a pretty good example from Nordic: https://devzone.nordicsemi.com/nordic/nordic-blog/b/blog/posts/extend-your-battery-life-with-nrf-cloud-coap-interface#mcetoc_1habr1j851

(there is also a specific section on power-savings, but it is more focused on MQTT vs. CoAP rather than CoAP w/ CID vs. CoAP w/o CID: https://devzone.nordicsemi.com/nordic/nordic-blog/b/blog/posts/extend-your-battery-life-with-nrf-cloud-coap-interface#mcetoc_1habr3tb55)

> What is the device use-case/scenario in which IP address changes? Isn't it like <1% of the possible cases? Personally I haven't notice IP address change at all and I can't even make up a scenario when that would happen. In case of cellular connectivity I guess that BTS roaming should solve that problem with maintaining single IP address (otherwise we would all be having connectivity problems when using mobile phones, where majority of connections/streams are TCP-based).

No, I don't think it is <1% of possible cases, though it is largely dependent on the network and use case. Using a cellular network is typically going to result in more frequent IP address changes, especially if you are in transit. You can observe this just by checking your phone's IP address periodically while driving or even walking around. You can also observe it my turning off cellular connectivity (e.g. airplane mode) and turning it back on.

> I'd like to see some evidence (at least theoretical) that backs that up. I've done some tests with qemu_x86 and lightdb/get sample, which shows that enabling CID support as done in this commit is increasing network usage (and as a result increasing power consumption). Below data shows almost 20% more data usage when sending empty CoAP messages, but there is almost 15% increase when issuing simple LightDB GET requests on /counter LightDB value.

I am working on collecting some more tangible data examples, but it is fundamentally true that record size is going to increase -- we are sending more data! Therefore, the question is whether the trade-off of not having to perform handshakes as frequently outweighs that cost of sending that extra data. In some cases it may and in others it may not. However, I think it is important to factor in the cases in which power savings are the most valuable. In the case of your laptop (or more broadly, a stationary device on a WiFi network that is communicating frequently), we might see net more data sent over the network because the handshakes are already infrequent and the application data is frequent. That being said, power savings are likely comparatively less critical in that domain, and network costs are lower (e.g. WiFi vs. cellular). On the other end of the spectrum, a device that communicates over cellular, is in transit, and sleeps often is more likely to benefit from the use of connection IDs _and_ is more likely to have tighter power constraints.

hasheddan (commented):
Noting here that most recent update does enable device -> cloud CID use:
image

net/golioth/Kconfig
sam-golioth: Per our discussion in office hours earlier today, we won't enable Connection ID support by default. We will write a lot of content to help clients understand when it makes sense to enable connection IDs in their own projects. I've updated this PR to remove the Kconfig changes.

hasheddan (approved):
Thanks @sam-golioth! LGTM 👍🏻

net/golioth/golioth.c
```diff
@@ -90,6 +90,18 @@ static int golioth_setsockopt_dtls(struct golioth_client *client, int sock,
 		}
 	}
 
+	/* If Connection IDs are enabled, set socket option to send CIDs, but not require that the
+	 * server sends one in return.
+	 */
+#ifdef CONFIG_MBEDTLS_SSL_DTLS_CONNECTION_ID
```
mniestroj, line 96: I think we should introduce new Kconfig option that will trigger whether Golioth connection will use DTLS CID or not. This option should select `CONFIG_MBEDTLS_SSL_DTLS_CONNECTION_ID` when enabled.

The main reason is that in some situation `CONFIG_MBEDTLS_SSL_DTLS_CONNECTION_ID=y` might be selected already (e.g. by another library/dependency) and in other user might want to use DTLS CID only for their purpose. Hence we should have a separate option that will control behavior of the Golioth SDK alone.

sam-golioth: Makes sense!

mniestroj (approved):

PR #447, opened 2023-10-31
ci: catch ram overflow build failures

sam-golioth:
Twister treats overflows in the RAM during builds as "skipped" tests instead of errors by default. Enabling the `--overflow-as-errors` option ensures we can catch overflows in CI.


mniestroj (approved):
Great catch!

I'd like to have the same flag enabled in `.gitlab-ci.ym`, but can do that later myself.

sam-golioth:
> Great catch!
> 
> I'd like to have the same flag enabled in `.gitlab-ci.yml`, but can do that later myself.

Added!

PR #449, opened 2023-12-21
OpenThread: avoid hardcoded Golioth IPv6 address

MarkoPura:
Avoid hardcoding the IPv6 Golioth System Server address by leveraging OpenThread DNS to retrieve the address before connecting to Golioth Cloud. As a side benefit, NAT64 prefix used by the Thred Border Router is set while synthesizing the address.

### DIFF
--- a/net/golioth/system_client.c
+++ b/net/golioth/system_client.c
@@ -17,7 +17,11 @@ LOG_MODULE_REGISTER(golioth_system, CONFIG_GOLIOTH_SYSTEM_CLIENT_LOG_LEVEL);
 #include <zephyr/sys/atomic.h>
 #include <zephyr/sys/util.h>
 
+#include "ot_dns.h"
+
 #define RX_BUFFER_SIZE		CONFIG_GOLIOTH_SYSTEM_CLIENT_RX_BUF_SIZE
+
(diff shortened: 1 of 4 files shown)


net/golioth/system_client.c
```diff
@@ -19,12 +19,42 @@ LOG_MODULE_REGISTER(golioth_system, CONFIG_GOLIOTH_SYSTEM_CLIENT_LOG_LEVEL);
 
 #define RX_BUFFER_SIZE		CONFIG_GOLIOTH_SYSTEM_CLIENT_RX_BUF_SIZE
 
+char golioth_system_server_host[40] = CONFIG_GOLIOTH_SYSTEM_SERVER_HOST;
```
sam-golioth, line 22: Let's make this `static`

net/golioth/system_client.c
```diff
@@ -244,11 +274,31 @@ SYS_INIT(golioth_system_init, APPLICATION,
 
 static int client_connect(struct golioth_client *client)
 {
-	int err;
+	int err = 0;
+
+#if defined (CONFIG_NET_L2_OPENTHREAD)
```
sam-golioth, line 279: Let's add a comment about why this extra step is necessary for Thread.

sam-golioth (approved):
Looks good!

PR #455, opened 2024-05-06
update recv_expiry when there is a network event

MarkoPura:
[As reported by a user](https://forum.golioth.io/t/lots-of-client-timeouts/581):
Client timeouts when there is no network traffic apart from sending a ping to the Server.

Having the value of `GOLIOTH_SYSTEM_CLIENT_PING_INTERVAL_SEC` smaller than the `GOLIOTH_SYSTEM_CLIENT_RX_TIMEOUT_SEC` results in a receive timeout, after which the connection is re-established.

### DIFF
--- a/net/golioth/system_client.c
+++ b/net/golioth/system_client.c
@@ -362,6 +362,7 @@ static void golioth_system_client_main(void *arg1, void *arg2, void *arg3)
 			(void)eventfd_read(fds[POLLFD_EVENT].fd,
 					   &eventfd_value);
 			LOG_DBG("Event in eventfd");
+			recv_expiry = k_uptime_get() + RECV_TIMEOUT;
 			event_occurred = true;
 		}
 


net/golioth/system_client.c
```diff
@@ -362,6 +362,7 @@ static void golioth_system_client_main(void *arg1, void *arg2, void *arg3)
 			(void)eventfd_read(fds[POLLFD_EVENT].fd,
 					   &eventfd_value);
 			LOG_DBG("Event in eventfd");
+			recv_expiry = k_uptime_get() + RECV_TIMEOUT;
```
EricNRS, line 365: From a short test, this fixes the timeout case when there is no other activity, so that is good.  From a code cleanliness standpoint line 399 is now redundant and should probably be removed.

mniestroj: Unfortunately following statement (in commit message):
> Update recv_expiry when there is a network event.

is not right. In theory `if (fds[POLLFD_EVENT].revents) {` should be entered **not** on network events, but on internal events in the SDK (like ping timeout, receive timeout, reconnect request, stop request, ...).

Not sure **why** that change helps, but it doesn't look right from the implementation perspective.

I still don't understand why the timeouts are happening, so I'll try to investigate that.

mniestroj: @EricNRS I can see why timeouts are not happening anymore with this patch. But are you sure that the connection is still valid? Are you sure that there is still some traffic between device and cloud when testing this change?

To me it looks like the internal timeout value (stored in `recv_expiry`) is updated and as a result there is no timeout triggered because of that. Pings are probably sent, but they might never reach cloud or there is no response to those pings.

EricNRS: @mniestroj - ah, I missed the fact that the first check is for the `POLLFD_EVENT` fds and not `POLLFD_SOCKET`.  Looking at firewall activity, I do see a connection to `coap.golioth.io` being established once per hour.  Since logging is currently disabled, I don't have any other way to confirm connectivity.

One interesting note is that with debug logging enabled in `system_client.c`, I do see the `Send PING` message, but I never see the `Received ping` message.

```[24:56:09.231,000] <dbg> golioth_system: golioth_system_client_main: Event in eventfd
[24:56:09.231,000] <dbg> golioth_system: golioth_system_client_main: Sending PING
[24:56:09.235,000] <dbg> golioth_system: golioth_system_client_main: Next timeout: 120000

[24:58:09.236,000] <dbg> golioth_system: golioth_system_client_main: Event in eventfd
[24:58:09.236,000] <dbg> golioth_system: golioth_system_client_main: Sending PING
[24:58:09.251,000] <dbg> golioth_system: golioth_system_client_main: Next timeout: 120000
```

mniestroj: > I do see the Send PING message, but I never see the Received ping message.

That is exactly what I suspected. In that case SDK should timeout, since there are no incoming messages. Since SDK behavior looks correct, the question is why no messages are received. This might be related to 120s ping being too big. I would try to use something smaller (e.g. 90s, 60s, 30s, ....) and observe results.

